PR #7, opened 2016-11-11
Use table instead of strcmp chain for modifier parsing

jpouellet:
Also adds support for `Mod2`

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2915,35 +2915,38 @@ static void load_default_config_values(Ghandles * g)
 static void parse_key_sequence(const char *seq, int *mask, KeySym * key)
 {
 	const char *seqp = seq;
+	char *name;
 	int found_modifier;
+	size_t i;
+
+	const struct {
+		char *name;
+		int mask;
+	} masks[] = {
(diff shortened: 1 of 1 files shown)


gui-daemon/xside.c
```diff
@@ -2915,36 +2915,36 @@ static void load_default_config_values(Ghandles * g)
...
-			found_modifier = 0;
-	} while (found_modifier);
+	for (i = 0; i < sizeof(masks)/sizeof(masks[0]); i++) {
+		name = masks[i].name;
+		if (strncasecmp(seqp, name, strlen(name)) == 0) {
+			*mask |= masks[i].mask;
+			seqp += strlen(name);
+			continue;
+		}
+		break;
```
marmarek: This will break parsing modifiers given in different order - for example will fail on `Alt-Ctrl-A` - because after parsing `Alt-` will not come back to check `Ctrl-`.

jpouellet: Still wrong, don't look yet ^_^.

jpouellet: Ah dammit. Yes. Should be good now.

PR #8, opened 2016-11-11
Implement "secure-pause" shortcut

jpouellet:
Ctrl-Alt-Shift-p pauses all VMs.

Partial solution to https://github.com/QubesOS/qubes-issues/issues/881


jpouellet:
Actually... I don't think this is the right place to implement such functionality. Closing now.

PR #12, opened 2017-02-21
Make watch-screen-layout-changes actually exit on SIGTERM

jaseg:
Previously, it would only exit on the next screen layout change after
SIGTERM.

This is solved by refactoring the code to use ```select``` on both X11's socket fd and a ```signalfd``` on ```SIGTERM```.


screen-layout-handler/watch-screen-layout-changes.c
```diff
@@ -40,39 +57,56 @@ int main(int argc, char **argv) {
...
+        FD_SET(sigfd, &in_fds);
+        FD_SET(x11_fd, &in_fds);
 
-        XNextEvent(d, &ev);
-        XRRUpdateConfiguration(&ev);
-
-        if (ev.type != xrr_event_base + RRScreenChangeNotify) {
-            /* skip other events (this shouldn't happen) */
-            continue;
+        if (select(FD_SETSIZE, &in_fds, NULL, NULL, NULL) < 0) {
```
marmarek: In theory the first `select` argument should be:
>        nfds  is the highest-numbered file descriptor in any of the three sets,
>        plus 1.

screen-layout-handler/watch-screen-layout-changes.c
```diff
@@ -40,39 +57,56 @@ int main(int argc, char **argv) {
...
-                exit(1);
-            case -1:
-                perror("fork");
-                exit(1);
-            default:
-                break;
+
+        while (XPending(d)) {
+            XNextEvent(d, &ev);
+            XRRUpdateConfiguration(&ev);
```
marmarek: Return value check?

screen-layout-handler/watch-screen-layout-changes.c
jaseg: You just confused me ;)

Yes, that's what the manpage says. [The manual](https://www.gnu.org/software/libc/manual/html_node/Waiting-for-I_002fO.html#index-select) has an example using ```FD_SETSIZE```, [the kernel](http://lxr.free-electrons.com/source/fs/select.c?v=4.10#L404) doesn't care and [this one book](https://books.google.de/books?id=kCTMFpEcIOwC&lpg=PA504&ots=zvLD_QVtsJ&dq=.&hl=de&pg=PA504#v=onepage&q&f=false) explains it all a bit saying its just a performance optimization and passing ```FD_SETSIZE``` is fine.

I'd leave this in since that call is only done once every couple of minutes at most.

screen-layout-handler/watch-screen-layout-changes.c
jaseg: As for ```XNextEvent```:
The return value [is not actually used](https://github.com/mirror/libX11/blob/2d20890e7ffd3ee88a9ceb25cdd2ac1fe7aaceb6/src/NextEvent.c#L37) neither is it specified in the manpage. Even if it were, we did just check there is an event in the queue using ```XPending```.

Regarding ```XRRUpdateConfiguration```:
True, that return value is not checked. I did not touch that as it has been this way. This is ok IMHO since that return value only indicates if ```XRRUpdateConfiguration``` did not understand an event, and we only selected events it understands, and [there is no harm calling it with one it doesn't](https://www.x.org/archive/X11R7.5/doc/man/man3/Xrandr.3.html#toc4) in any case.

PR #21, opened 2018-02-16
R4.1: Support new GUI protocol version with flexible MSG_WINDOW_DUMP

HW42:
This replaces the now deprecated MSG_MFNDUMP. For now the there's only
MSG_WINDOW_DUMP_TYPE_GRANT_REFS which uses grant references instead of
plain mfns as before.


HW42:
Depends on https://github.com/QubesOS/qubes-gui-common/pull/2

HW42:
The Travis builds fails because it misses the new header from https://github.com/QubesOS/qubes-gui-common/pull/2 (same for https://github.com/QubesOS/qubes-gui-agent-linux/pull/34)

shmoverride/shmoverride.c
```diff
@@ -32,91 +32,208 @@
...
+struct info {
+    uint32_t type;
+    union {
+        struct mfns_info mfns;
+        struct grant_refs_info grant;
+    } u;
+};
+
+static void shmat_mfns(struct shm_args_hdr *shm_args, uint8_t **map,
+                       struct info *info) {
```
marmarek: Better have it return mapped pointer (same as `shmat`) instead of `map` output argument.

shmoverride/shmoverride.c
```diff
@@ -32,91 +32,208 @@
...
+    *map = xc_map_foreign_pages(xc_hnd, shm_args->domid, PROT_READ,
+                                pfntable, shm_args_mfns->count);
+    if (*map == NULL)
+        return;
+
+    *map += shm_args_mfns->off;
+}
+
+static void shmat_grant_refs(struct shm_args_hdr *shm_args, uint8_t **map,
+                             struct info *info) {
```
marmarek: Same here.

shmoverride/shmoverride.c
```diff
@@ -236,20 +359,20 @@ int __attribute__ ((constructor)) initfunc()
...
+    shm_args->shmid = local_shmid;
     return 0;
 }
 
 int __attribute__ ((destructor)) descfunc()
 {
-    if (cmd_pages) {
-        real_shmdt(cmd_pages);
+    if (shm_args) {
+        real_shmdt(shm_args);
```
marmarek: While not strictly required at this point, better call also `xengnttab_close`.

HW42:
Do you prefer followup commits for improvements or replacing the existing commit with a improved one? I prefer the later since it leads to a clean history but it makes reviewing harder since you need to track old versions yourself.

marmarek:
I'm ok with either.

gui-daemon/xside.c
```diff
@@ -2358,8 +2375,158 @@ static void handle_mfndump(Ghandles * g, struct windowdata *vm_window)
...
+           && (int) untrusted_wd_hdr.height <= MAX_WINDOW_HEIGHT);
+    vm_window->image_width = untrusted_wd_hdr.width;
+    vm_window->image_height = untrusted_wd_hdr.height;
+    //VERIFY(untrusted_wd_hdr.bpp == 24);
+
+    handle_window_dump_body(g, wd_type, untrusted_wd_body_len, &img_data_size,
+                            &shm_args, &shm_args_len);
+
+    if (g->invisible)
+        return;
```
marmarek: shm_args is leaked.

shmoverride/shmoverride.c
```diff
@@ -32,91 +32,212 @@
...
+            shm_args->domid,
+            &shm_args_grant->refs[0],
+            PROT_READ);
+
+    if (map != NULL)
+        return map;
+
+    // Something failed. Most likely the other domain already destroyed the
+    // buffer and thereby invalidated the refs. So create a dummy buffer. This
+    // mapping will probably be unmapped very soon anyway.
```
marmarek: Why return dummy buffer, instead of a failure (NULL)? How grant refs version is special here?

marmarek: It's about avoiding gui-daemon crash:
```
ErrorHandler: BadAccess (attempt to access private resource denied)
                 Major opcode: 130 (MIT-SHM)
                 Minor opcode: 1 (X_ShmAttach)
                 ResourceID:   0x1f5
                 Failed serial number:  101
                 Current serial number: 102
```
And from shmoverride:
```
gnttab: error: mmap failed: Invalid argument
```

I'm looking into more graceful handling of this error. I still think allocating (potentially large) memory buffer just to mute an error is wrong.

marmarek: And BTW this isn't specific to grant refs, it can also happen with old method: https://github.com/QubesOS/qubes-issues/issues/4071

marmarek: The problem with this is `XShmAttach` still reports success, but then the error is reported during subsequent `XSync` call. But since we need that `XSync` call anyway and I get detailed error info, I can correlate those two things (using global variable :/). This will fix QubesOS/qubes-issues#4071 too.

PR #23, opened 2018-07-27
Correct note about "temporary" shm lifecycle

jpouellet:
The description sounds like a reasonable way to implement it, but it's apparently not how it's been implemented. I'm not sure why, but one theoretical reason might be to preemtively avoid potential bugs related to shmid recycling. For example, since X (allegedly) internally keeps track of shmids, if the temporary shmid matches one previously seen for a still-existing window, one might imagine bugs where the windows might end up with the same backing instead of X calling through to shmoverride.

In any case, this is confirmable by reading the code (look for `shmctl(..., IPC_RMID, ...)`), and by watching `/proc/sysvipc/shm`.

### DIFF
--- a/shmoverride/README
+++ b/shmoverride/README
@@ -18,7 +18,9 @@ and from which domain.
 	Somewhat unfortunately, the Xorg server tracks the already attached shmids.
 Therefore, it is not possible to pass the same "magic" value of synth_shmid
 to XShmAttach(...synth_shmid...). Before each XShmAttach, qubes_guid creates
-a temporary ("real") shared memory segment, sets cmd_pages->shmid to it, and
-then executes XShmAttach. This temporary segment is destroyed after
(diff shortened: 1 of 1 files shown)


marmarek:
> I'm not sure why, but one theoretical reason might be to preemtively avoid potential bugs related to shmid recycling. For example, since X (allegedly) internally keeps track of shmids, if the temporary shmid matches one previously seen for a still-existing window

This is exactly the reason, see https://github.com/QubesOS/qubes-gui-daemon/commit/2d3e090b9def6dc5914d7e8ca8c76a7553cebc37#diff-035a767b7d35dbcd864a2b5be4770c6c
AFAIR this happened to me at least once.

jpouellet:
I see, thanks. Force-pushed a better explanation.

PR #28, opened 2019-07-16
Add README

tapaswenipathak:
Ref https://github.com/QubesOS/qubes-issues/issues/4260.


README.md
```diff
@@ -0,0 +1,22 @@
+qubes-gui-daemon
+================
+
+gui-daemon is Qubes GUI Virtualisation agent/daemon, present on Dom0 side.
+
+gui-daemon has default daemon setting. Starting the GUI will ask you whether to
+start the daemon and you can then accept. The GUI domain would be the same as
```
marmarek: Can you explain what you mean by "the GUI will ask you whether to start the daemon and you can then accept"? There is no interactive prompt, just command like arguments.

README.md
```diff
@@ -0,0 +1,22 @@
...
+admin domain in all but few scenarios. With just few `#ifdefs` the existing
+GUI daemon should compile and run on most OS from Linux/Xorg, Windows and MAcs.
```
marmarek: If you mean running gui-daemon on non-Xorg system, this is not true. Much of the code is very much Xorg specific. It is definitely possible to add support for other OSes, but unfortunately it isn't "just few `#ifdefs`".

README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The config is defined by guid.conf.
+
+![](./assets/qubes-components.png)
```
marmarek: It may be also good to link to https://www.qubes-os.org/doc/gui/

README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Building qubes-gui-daemon
+
+make `gui-daemon`
```
marmarek: This is true, but an important detail is missing: it should be run from qubes-builder, not this repository directly. https://www.qubes-os.org/doc/qubes-builder/

Alternatively `make all BACKEND_VMM=xen` here. But will require installing dependencies manually.

tapaswenipathak:
@marmarek how do you suggest adding me the qubes-components image? What do you have in mind?

README.md
tapaswenipathak: From my observation you can have the enablement if you use libvirt.

marmarek:
> @marmarek how do you suggest adding me the qubes-components image? What do you have in mind?

Linking to https://www.qubes-os.org/attachment/wiki/posts/qubes-components.png

tapaswenipathak:
@marmarek looks ok?

tapaswenipathak:
@marmarek looks ok?

README.md
marmarek: This comment still apply

README.md
marmarek: It is possible to add support for other operating systems. But the code as it is right now supports only Linux/Xorg.

marmarek:
While Whonix do use mediawiki format, we try to not add other formats than markdown (`.md`) and in some places reStructuredText (`.rst`). So, please convert to markdown.

README.md
tapaswenipathak: Could you write the statement properly?

tapaswenipathak:
@marmarek Just curious have you implemented any `.mediaiwiki` file formatting before? Just as an FYI there is no use of having `.mediawiki` formatting if you use `.md` formatting, sample commit cb9c52d. Why not have only `.md` then? What is the use of changing the file formatting as per `.mediawiki` rules?

PR #29, opened 2019-09-10
Add keypress initiated wipe of VM clipboard

johanna-a:
When using a password manager in an isolated VM and copying passwords from
there, it is convenient to have a quick method to wipe the clipboard of the
active VM after the password has been used. This adds the keyboard shortcut
Ctrl+Shift+B (for blank) as default to paste a single space into the VM
clipboard.

Note that this is a proof-of-concept/request-for-comments at this stage, not including support for
stubdom VMs nor having been thoroughly tested. Also, actually blanking the clipboard instead of replacing it with a single space would be nicer. But I wanted to see if there's any interest in this feature before going forward with development :)

Also note that this has been tested against r4.0 with the fc25 dom0, but not (yet) on fc29 dom0.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -875,6 +875,41 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
 
         return 1;
     }
+
+    /* blank */
+    if (((int)ev->state & SPECIAL_KEYS_MASK) == g->blank_seq_mask
+        && ev->keycode == XKeysymToKeycode(g->display, g->blank_seq_key)) {
+        if (ev->type != KeyPress)
+            return 1;
+        inter_appviewer_lock(g, 1);
+        clipboard_file_xevent_time = get_clipboard_file_xevent_timestamp();
+        if (clipboard_file_xevent_time > ev->time) {
+            /* some other clipboard operation happened in the meantime, discard
+             * request */
+            inter_appviewer_lock(g, 0);
+            fprintf(stderr,
+                    "received clipboard xevent after some other clipboard op, discarding\n");
+            return 1;
+        }
+        if (g->qrexec_clipboard) {
+            if (g->log_level > 0)
+                fprintf(stderr, "secure blank is not supported for stubdom GUI\n");
+        } else {
+            hdr.type = MSG_CLIPBOARD_DATA;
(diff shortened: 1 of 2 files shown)


johanna-a:
I've uploaded my key to the sks public keyservers but it seems it has not been added yet. Just FYI.

johanna-a:
> Have you tried sending empty clipboard message, with untrusted_len=0? By code inspection of gui-agent-linux, it should work.
You are right, untrusted_len=0 does work. I don't know where but I thought I followed that execution path and found a malloc(len) somewhere but that might be when I was still chasing the stubdom code. I updated with an empty clipboard message now, the code can really use a cleanup though. I'm still considering this unfinished before it has stubdom support and perhaps also support for some notification that the clipboard is blanked (that seems non-trivial though)
> There is also related discussion here: [QubesOS/qubes-issues#3415](https://github.com/QubesOS/qubes-issues/issues/3415)
> The idea there is to do it automatically after delay, instead of key press.
I don't think these ideas are mutually exclusive. I like the control of initiating clearing the clipboard with the press of a button, but someone else may prefer a timeout, or both. My idea for adding a timeout is that clearing the clipboard will be dom0 initiated depending on a flag on the VM that the clipboard data come from. That would enable passwords copied from a vault to be cleared while url:s copied form other VMs would not. This isn't perfect with regards to the not clearing if the clipboard data has already been modified, but it might be good enough. The solution to that is probably introducing a new command (or the special case of clipboard data of length 0) that clears the clipboard if it hasn't been changed since the last qubes clipboard operation. As for initiating the clear from the password vault VM, I think that would just be too messy.

marmarek:
No activity, and https://github.com/QubesOS/qubes-issues/issues/3415 got implemented in the meantime. If you still want to do this one, please rebase the code and open new PR (or force-push here and leave a comment).

johanna-a:
Sorry for the late reply.
Due to the lack of interest in this particular feature, especially compared to the automatic keyboard wipe implemented in QubesOS/qubes-issues#3415, I didn't push for this. I still personally need this feature, but I've implemented it locally in a way that is easy enough to maintain which works for me.

Should anyone else want the possibility of manually, via keypress, initiating a wipe of the VM clipboard, I'd be happy to pick this work up in the future.

PR #32, opened 2019-10-20
Drop set of /qubes-keyboard as it's done core GUI extension

fepitre:


marmarek:
Templates made on R4.0 should still work on R4.1, so this can't be dropped yet. Only in R5.0. I'll close it for now, as it will probably conflict with other changes until then anyway.

PR #35, opened 2020-02-27
Audiovm

fepitre:
Related PR QubesOS/qubes-core-admin-client/pull/118
https://github.com/QubesOS/qubes-issues/issues/1590

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2837,60 +2837,6 @@ static void release_all_mapped_mfns(void)
     }
 }
 
-/* start pulseaudio Dom0 proxy */
-static void exec_pacat(Ghandles * g)
-{
-    int i, fd, maxfiles;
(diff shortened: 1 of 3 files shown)


marmarek:
The missing piece is getting this separate packages (qubes-audio-daemon, qubes-audio-dom0) installed. qubes-audio-dom0 should be installed in dom0 in all the cases.
But qubes-audio-daemon should be optional (to be installed only in AudioVM). I mean, not necessary avoid installing it, but make it possible to remove, without removing other important packages (which means adding `R: qubes-audio-daemon` to some key dom0 package is a bad idea).
To make things even more complex, this should be handled also during R4.0->R4.1 upgrade.

Maybe it's time for meta-package(s) in dom0 too?

fepitre:
Yes I would vote for dom0 meta packages.

PR #37, opened 2020-03-14
Convert qubes.WindowIconUpdater to a socket service

pwmarcz:
As a result, we have a single service running for all the VMs.

Start it using /etc/xdg/autostart to ensure it's running in the
proper graphical session.

Depends on Image.get_from_stream_async, implemented in
QubesOS/qubes-linux-utils#52.

### DIFF
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -117,8 +117,10 @@ class IconReceiver(object):
 
     def get_color(self, domain):
         # Load the VM properties - we need this only to get VM color
-        vm = self.app.domains[domain]
-        if vm is None:
+        self.app.domains.refresh_cache(force=True)
+        try:
+            vm = self.app.domains[domain]
+        except KeyError:
             raise QubesException("VM '{}' doesn't exist in qubes.xml".format(
                 domain))
         return vm.label.color
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -27,11 +27,14 @@ import sys
 import asyncio
 import argparse
 from qubesimgconverter import ICON_MAXSIZE, Image
+import logging
 
 import xcffib as xcb
 import xcffib.xproto as xproto
 from qubesadmin import Qubes
 from qubesadmin.exc import QubesException
+
+log = logging.getLogger('icon-receiver')
 
 BYTES_PER_PIXEL = 4
 BITS_PER_PIXEL = 32
@@ -349,7 +352,7 @@ class IconReceiver(object):
             domain = header_parts[1]
             color = self.get_color(domain)
 
-            print('connected: {}'.format(domain), file=sys.stderr)
+            log.info('connected: %s', domain)
 
             while True:
                 untrusted_w = await reader.readline()
@@ -366,7 +369,7 @@ class IconReceiver(object):
                 except KeyError:
                     self.cache_icon(remote_winid, icon_property_data)
 
-            print('disconnected: {}'.format(domain), file=sys.stderr)
+            log.info('disconnected: %s', domain)
         finally:
             writer.close()
             await writer.wait_closed()
@@ -390,6 +393,10 @@ def main():
                   '(run with --force to override)',
                   file=sys.stderr)
             return
+
+    logging.basicConfig(
+        stream=sys.stderr, level=logging.INFO,
+        format='%(asctime)s %(name)s: %(message)s')
 
     rcvd = IconReceiver()
 
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -22,9 +22,10 @@
 #
 #
 import os
-import select
 import struct
 import sys
+import asyncio
+import argparse
 from qubesimgconverter import ICON_MAXSIZE, Image
 
 import xcffib as xcb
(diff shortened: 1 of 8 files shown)


pwmarcz:
Ouch, this is not right, I think the cache needs to depend on domain ID...

marmarek:
Not only cache, but also window ID map. Or at least the map should now have two fields (vm, remote window id). IMO it would be cleaner to have separate object instances for each incoming connection and keep domain-specific data there.
A little issue could be with dispatching local X events to the right instance, but we do already retrieve local window -> VM name mapping (search for `atom_vmname`), so it shouldn't be much of an issue.

pwmarcz:
What do you think about the current code? I converted the "remote ID" to `(domain_name, integer)` tuple. This takes care of both cache and mapping.

I think creating separate object instances would make it less clean, because the X server connection is global and not related to either a specific domain or connection. You would need to filter X messages (which we receive globally) down to the right instance, and handle creating/destroying these instances.

With the current code, we store very little state per connection (just the local variables: domain and color), and we keep open the possibility of multiple connections from icon-sender, for instance per-request instead of per-session (that's not to say we should do that, but I think it shows things are more decoupled).

window-icon-updater/icon-receiver
```diff
@@ -331,32 +323,90 @@ class IconReceiver(object):
...
-            if remote_fd in read_fds:
-                untrusted_w = sys.stdin.buffer.readline(32)
+    async def handle_client(self, reader, writer):
+        try:
+            # Parse header from qrexec
+            header = await reader.readuntil(b'\0')
+            header_parts = header.decode('ascii').split(' ')
+            assert len(header_parts) >= 2, header_parts
+
+            service_name = header_parts[0].rstrip('+')
```
marmarek: Why rstrip? Service name is until first `+`, not last.

marmarek:
Ok, makes sense.

window-icon-updater/icon-receiver
```diff
@@ -331,32 +323,90 @@ class IconReceiver(object):
...
                     self.cache_icon(remote_winid, icon_property_data)
 
-if __name__ == '__main__':
+            print('disconnected: {}'.format(domain), file=sys.stderr)
+        finally:
+            writer.close()
+            await writer.wait_closed()
+
+
+def get_color(domain):
```
marmarek: Why outside of the class?
Now when it could be called multiple times, it would make sense to keep the `Qubes()` object, not construct it every time.

window-icon-updater/icon-receiver
pwmarcz: Rewritten to indicate the intention better (we expect `qubes.WindowIconUpdater` and empty argument).

window-icon-updater/icon-receiver
pwmarcz: Right. Moved `self.app = Qubes()` to constructor.

pwmarcz:
By the way, master is still broken for me:

```
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]: Traceback (most recent call last):
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:   File "/usr/lib/qubes/icon-receiver", line 361, in <module>
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:     rcvd = IconReceiver()
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:   File "/usr/lib/qubes/icon-receiver", line 72, in __init__
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:     self.conn = xcb.connect()
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:   File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 526, in __init__
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:     self.invalid()
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:   File "/usr/lib/python3.7/site-packages/xcffib/__init__.py", line 558, in invalid
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]:     raise ConnectionException(err)
Mar 18 20:20:39 dom0 qubes.WindowIconUpdater+-work[7561]: xcffib.ConnectionException: xcb connection errors because of socket, pipe and other stream errors.
```

marmarek:
Does it happen after logoff+logon?

pwmarcz:
No, right after reboot.

marmarek:
It might be because of the qrexec-policy-daemon. Since it's now this service calling qrexec-client, not qrexec-daemon, the actual service is called a) as root, b) with limited environment variables.

marmarek:
Since https://github.com/QubesOS/qubes-core-qrexec/pull/41 is merged, I assume this should make use of it, right?

marmarek:
This also needs adjusted dependencies, for all the new things it requires:
 - qubes-core-qrexec >= 4.1.5 (socket services)
 - qubes-utils >= 4.1.4 (async image)

Both Fedora and Debian.

window-icon-updater/icon-receiver
```diff
@@ -243,13 +234,24 @@ class IconReceiver(object):
...
+    async def handle_events(self):
+        # Emulate select()
+        event = asyncio.Event()
+        asyncio.get_event_loop().add_reader(
+            self.conn.get_file_descriptor(), event.set)
+
+        while True:
+            await event.wait()
+            event.clear()
+            self.handle_pending_events()
```
marmarek: Why not simply connect handle_pending_events() as a reader function directly?

pwmarcz: Right. Updated.

pwmarcz:
Ouch, there is one more edge case: creating a new machine while the process is running :)

PR #38, opened 2020-03-18
gui-daemon: set qubes-keyboard only in dom0

fepitre:


gui-daemon/xside.c
```diff
@@ -3435,7 +3435,7 @@ int main(int argc, char **argv)
     struct stat stat_buf;
     char *display_str;
     int display_num;
-
+    char* qubes_release = "/etc/qubes-release";
```
marmarek: This should be `#define`, in `xside.h`.

PR #39, opened 2020-03-24
shmoverride: handle Xorg crash leaving shm.id file

pwmarcz:
Instead of opening with O_CREAT|O_EXCL, use flock() to see if
the file is held by an alive Xorg process.

See 0ac0c7fcee1dcbc6235718b1b1fcf90131811973 for more context.

Solves part of QubesOS/qubes-issues#5273 (qubes-guid failing to
start after Xorg crash).

### DIFF
--- a/shmoverride/shmoverride.c
+++ b/shmoverride/shmoverride.c
@@ -37,6 +37,8 @@
 #include <alloca.h>
 #include <errno.h>
 #include <unistd.h>
+#include <sys/file.h>
+#include <assert.h>
 #include "list.h"
 #include "shm-args.h"
 #include <qubes-gui-protocol.h>
@@ -57,6 +59,7 @@ static xengnttab_handle *xgt;
 static int list_len;
 static char __shmid_filename[SHMID_FILENAME_LEN];
 static char *shmid_filename = NULL;
+static int idfd = -1;
 static char display_str[SHMID_DISPLAY_MAXLEN+1] = "";
 
 struct mfns_info {
@@ -226,7 +229,7 @@ int shmctl(int shmid, int cmd, struct shmid_ds *buf)
     return 0;
 }
 
-void get_display()
+int get_display()
 {
     int fd;
     ssize_t res;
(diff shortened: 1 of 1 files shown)


shmoverride/shmoverride.c
```diff
@@ -319,12 +322,32 @@ int __attribute__ ((constructor)) initfunc()
...
+     * process might have crashed).
+     */
+    idfd = open(shmid_filename, O_WRONLY | O_CREAT, 0600);
     if (idfd < 0) {
-        fprintf(stderr, "shmoverride creating %s: %s\n",
+        fprintf(stderr, "shmoverride opening %s: %s\n",
             shmid_filename, strerror(errno));
-        xc_interface_close(xc_hnd);
-        return 0;
+        exit(1);
```
marmarek: Exit here is a bad idea. This is preloaded into X server, so it will terminate the whole X server, even if shmoverride wouldn't be really needed (installer, firstboot, X -configure, ...).

shmoverride/shmoverride.c
```diff
@@ -319,12 +322,32 @@ int __attribute__ ((constructor)) initfunc()
...
+        fprintf(stderr, "shmoverride flock %s: %s\n",
+                shmid_filename, strerror(errno));
+        if (errno == EWOULDBLOCK) {
+            /* There is an alive process holding the file, give up. */
+            close(idfd);
+            idfd = -1;
+            xc_interface_close(xc_hnd);
+            return 0;
+        }
+        exit(1);
```
marmarek: Again, please no exit.

shmoverride/shmoverride.c
```diff
@@ -319,12 +322,32 @@ int __attribute__ ((constructor)) initfunc()
...
+    }
+    if (ftruncate(idfd, 0) < 0) {
+        perror("shmoverride ftruncate");
+        unlink(shmid_filename);
+        exit(1);
```
marmarek: This one is unlikely to happen, but still - please change to return.

pwmarcz:
Okay. There were a few more exits in the existing code, and it looked like in all cases we could clean up and continue, so I changed them.

PR #40, opened 2020-03-26
Support MSG_CURSOR

pwmarcz:
See QubesOS/qubes-issues#1551.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -61,6 +61,17 @@
 #include "trayicon.h"
 #include "shm-args.h"
 
+/* Supported protocol version */
+
+#define PROTOCOL_VERSION_MAJOR 1
+#define PROTOCOL_VERSION_MINOR 3
+#define PROTOCOL_VERSION (PROTOCOL_VERSION_MAJOR << 16 | PROTOCOL_VERSION_MINOR)
+
+#if !(PROTOCOL_VERSION_MAJOR == QUBES_GUID_PROTOCOL_VERSION_MAJOR && \
+      PROTOCOL_VERSION_MINOR <= QUBES_GUID_PROTOCOL_VERSION_MINOR)
+#  error Incompatible qubes-gui-protocol.h.
+#endif
+
 /* some configuration */
 
 static Ghandles ghandles;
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -459,6 +461,21 @@ static void mkghandles(Ghandles * g)
...
+    if (!g->cursors) {
+        perror("malloc");
+        exit(1);
+    }
+    for (i = 0; i < XC_num_glyphs; i++) {
+        /* X font cursors have even numbers from 0 up to XC_num_glyphs.
+         * Fill the rest with None.
+         */
+        if (i % 2 == 0)
+            g->cursors[i] = XCreateFontCursor(g->display, i);
```
marmarek: Can this return an error? Or will it just call `x11_error_handler()`, which will abort the whole thing?

pwmarcz: I just tested that.

The function is happy to always return a newly-allocated value (it looks like it's allocated in library before the X server validates the request). However, using an out-of-range value generates `BadValue`, and as soon as `x11_error_handler()` is installed, it aborts the program. (The handler is installed just after executing this function, but it still catches the errors generated before). I assume `BadAlloc` will behave in a similar way.

Providing an in-range value not corresponding to any cursor (odd numbers) does not fail, but the cursor still behaves as None, i.e. a default cursor is used instead. (This shouldn't happen but it's nice to see that there is no "invisible cursor" failure mode).

pwmarcz:
Apparently [no version numbers may be used with virtual packages](https://rpm.org/user_doc/dependencies.html), at least in RPM?

But I think static assertions would be good enough. So how about using a code like this, in both gui-agent and gui-daemon:

```c
#define PROTOCOL_VERSION_MAJOR 1
#define PROTOCOL_VERSION_MINOR 3
#define PROTOCOL_VERSION (PROTOCOL_VERSION_MAJOR << 16 | PROTOCOL_VERSION_MINOR)

#if !(PROTOCOL_VERSION_MAJOR == QUBES_GUID_PROTOCOL_VERSION_MAJOR && \
      PROTOCOL_VERSION_MINOR <= QUBES_GUID_PROTOCOL_VERSION_MINOR)
#  error Incompatible qubes-gui-protocol.h.
#endif
```

PR #44, opened 2020-04-30
set override redirect attribute on window creating

jevank:
It is not necessary with xfwm, but is critical with Compiz because of event racing from decoration plugin and guest side.


marmarek:
What exactly is the issue with Compiz? Window manager shouldn't do anything about the window before it is mapped (XMapWindow call), and the override redirect flag is set before window gets mapped (see handle_map).

jevank:
Without this patch Compiz time to time decorates Menu and Pop-up windows. With Pop-up windows it is often brings to blinking window and even crush window manager. As I understood debug messages and source codes, Compiz decoration plugin force decoration right after window creating and before it mapping (if override redirect flag not set). It causes configure event to window and guest configure messages drop until wait for acknowledgement.

PR #45, opened 2020-06-17
QI#2624: enable rec_requested by default

easydozen:
Sometimes pacat-simple-vchan starts too late to catch
QUBES_PA_SOURCE_START_CMD and rec_requested flag stays turned off.
As result sound recording doesn't work.

QubesOS/qubes-issues#2624


marmarek:
Hmm, I don't see how late pacat-simple-vchan start could cause this. Vchan is a ring-buffer based communication - even if the client (pacat-simple-vchan) connects late, it can retrieve buffered messages.

What I think could happen here is module-vchan-sink not sending QUBES_PA_SOURCE_START_CMD message at all. See `source_set_state_in_io_thread_cb` - it reacts to a state change. Maybe if the source starts in PA_SOURCE_RUNNING state initially, the message isn't sent?

easydozen:
if pacat-simple-vchan starts as soon as possible then sound record is always available on my tests. 

I''m going to investigate it further.

PR #48, opened 2020-06-28
fix audio recording issue

easydozen:
QubesOS/qubes-issues#2624

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -417,14 +417,18 @@ static void stream_state_callback(pa_stream *s, void *userdata) {
 	    if (u->play_stream == s) {
 		u->play_stdio_event = u->mainloop_api->io_new(u->mainloop_api,
                     libvchan_fd_for_select(u->play_ctrl), PA_IO_EVENT_INPUT, vchan_play_callback,  u);
-		if (!u->play_stdio_event)
+		if (!u->play_stdio_event) {
 		    pacat_log("io_new play failed");
+		    quit(u, 1);
+		}
 	    }
 	    if (u->rec_stream == s) {
 		u->rec_stdio_event = u->mainloop_api->io_new(u->mainloop_api,
 			libvchan_fd_for_select(u->rec_ctrl), PA_IO_EVENT_INPUT, vchan_rec_callback, u);
-		if (!u->rec_stdio_event)
+		if (!u->rec_stdio_event) {
 		    pacat_log("io_new rec failed");
+		    quit(u, 1);
+		}
 	    }
             break;
 
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -307,7 +307,6 @@ static void vchan_play_callback(pa_mainloop_api *UNUSED(a),
     }
 
     /* process playback data */
-    if (u->play_stream && pa_stream_get_state(u->play_stream) == PA_STREAM_READY)
         process_playback_data(u, u->play_stream, pa_stream_writable_size(u->play_stream));
 }
 
@@ -325,7 +324,6 @@ static void vchan_rec_callback(pa_mainloop_api *UNUSED(a),
         return;
     }
 
-    if (u->rec_stream && pa_stream_get_state(u->rec_stream) == PA_STREAM_READY) {
         /* process VM control command */
         uint32_t cmd;
         while (libvchan_data_ready(u->rec_ctrl) >= (int)sizeof(cmd)) {
@@ -370,7 +368,6 @@ static void vchan_rec_callback(pa_mainloop_api *UNUSED(a),
         if (libvchan_buffer_space(u->rec_ctrl)) {
             send_rec_data(u->rec_stream, u);
         }
-    }
 }
 
 /* This routine is called whenever the stream state changes */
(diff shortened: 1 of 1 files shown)


marmarek:
Assume this should also fix the issue the tests detected, right?

marmarek:
There is a little (I guess unintentional) behavior change - now when `io_new()` fails, pacat process isn't terminated, it only logs an error. That shouldn't be much of a problem, since I have never seen this failure actually happening. But it would be nice to fix it.

easydozen:
> Assume this should also fix the issue the tests detected, right?

There are two issues that detected by tests: not working audio recording and sound distortion. 
Here we are dealing with the first one, that leads to "only silence detected" error.

Another one - the sound distortion issue that shown as "frequency not in range" error. I just don't have enough test results to be sure how to fix it in a right way. It could be tackled in stubdom init with a proper audio parameters selection, but we will have to hardcode them and stick with selected default audio model. 
It would be great to hear your thoughts.

marmarek:
> It could be tackled in stubdom init with a proper audio parameters selection, but we will have to hardcode them and stick with selected default audio model.
> It would be great to hear your thoughts.

There is also an option to add those parameters depending on selected audio model - add custom parameters only when certain audio model is selected, and keep defaults otherwise.
Anyway, as you can see in the test results, much more common case is silence, not distortion.

marmarek:
Strangely enough, if I get the exact disk image tests were using and run those audio tests again, I can't reproduce silence, but sometimes I get "frequency not in specified range".

easydozen:
> There is also an option to add those parameters depending on selected audio model - add custom parameters only when certain audio model is selected, and keep defaults otherwise.

Good Idea. I'll make ich6 preset for now.

> Anyway, as you can see in the test results, much more common case is silence, not distortion.

For silence issue it's enough to have:
- preinstalled alsa-utils in template - add alsa-utils dependency in one of qubes domU packages;
- unmuted sink at first boot - run pacmd set-sink-mute and pacmd set-sink-volume in hvm tests.

Would you prefer to have a more elegant solution here?

> Strangely enough, if I get the exact disk image tests were using and run those audio tests again, I can't reproduce silence, but sometimes I get "frequency not in specified range".

Problem that you are facing might be explained by save-restore audio settings. At first boot sink was muted ("silence detected"), then manually changed settings was saved by pulseaudio/alsa. Next test execution ends up with "frequency not in range".

marmarek:
> preinstalled alsa-utils in template - add alsa-utils dependency in one of qubes domU packages;

It is in the template already ([for future templates added here](https://github.com/QubesOS/qubes-builder-debian/commit/374f84f585f91646b2f82495761c0562e4e10d30) and also added via salt in tests preparation).

> unmuted sink at first boot - run pacmd set-sink-mute and pacmd set-sink-volume in hvm tests.

Unlikely to be the reason specifically in tests - there should be no state leaking between failed and successful attempts (appvm is created each time fresh and template wasn't started in the between).

> Problem that you are facing might be explained by save-restore audio settings. At first boot sink was muted ("silence detected"), then manually changed settings was saved by pulseaudio/alsa.

I've just re-run this particular test job. It started from exactly the same initial disk image (no state from previous run). And this time it worked better (see updated comment above). Perhaps it's about some race condition, manifesting itself when VM is slower? During the failed attempt(s) there was higher load on the physical host (about 3 VMs with those tests running, instead of one now).

easydozen:
> I've just re-run this particular test job. It started from exactly the same initial disk image (no state from previous run). And this time it worked better (see updated comment above). Perhaps it's about some race condition, manifesting itself when VM is slower? During the failed attempt(s) there was higher load on the physical host (about 3 VMs with those tests running, instead of one now).

I've tried to run hvm tests several times on heavy loaded system (2 VMs with stress -c 10 -i 10) - nothing suspicious. May I take your image somehow?

marmarek:
Sure, you can download it from https://openqa.qubes-os.org/tests/10412#downloads
See also autoinst-log.txt for qemu cmdline (you need a KVM host...).

easydozen:
I've discovered the strange thing that you're talking. I tried to run tests like described in wiki with qubesd stop/start commands. In this case the problem is not reproduced.

It seems that pacat-simple-vchan doesn't start at every second domain-name run. 
systemctl restart qubesd or manually starting qvm-start-daemon fixes daemon startup. 

I'm seeking for a root of this problem (event handlers?). If issue seems familiar, please, push me to the right way.

easydozen:
On failed attempt pacat-simple-vchan writes the following:
domain dead
libvchan_client_init: No such file or directory

Meanwhile QUBES_PA_SINK_VCHAN_PORT and QUBES_PA_SOURCE_VCHAN_PORT are available in xenstore.

marmarek:
Is the domain ID given to pacat-simple-vchan correct? Maybe it's stale value, or even -1?

easydozen:
> Is the domain ID given to pacat-simple-vchan correct? Maybe it's stale value, or even -1?

Yes, you're right! xid stays at first run value.

marmarek:
Ok, then I think I've missed this one in https://github.com/QubesOS/qubes-core-admin/pull/346

marmarek:
Now, with QubesOS/qubes-core-admin#358, it is significantly better. Some tests are still running (the summary above will update only after all finishes), but you can already view results here: https://openqa.qubes-os.org/tests/10422
There is one silence (with whonix-ws-15 template), one "freq not in range" error and one timeout on startup. In all the cases on "audio_rec_unmuted_hvm" test (but one time it passed). Audio playback works in all the cases.

easydozen:
> There is one silence (with whonix-ws-15 template), 
I'll explicitly set volume in case it's still save/restore issue.

> one "freq not in range" error
Registered frequency below original value likely means that a part of audio data wasn't recorded by some reason. Above - noisy environment. 
I think audio model preset should fix this.
 
> one timeout on startup. 
Increased timeout in wait_pulseaudio_startup may help.

marmarek:
> > one timeout on startup.
> 
> Increased timeout in wait_pulseaudio_startup may help.

Oh, I see it is a timeout of `qubes.WaitForSession` service call - so a generic VM startup timeout, not specifically pulseaudio.

easydozen:
Another thing that we should pay attention for - pacat-simple-vchan remains running after test execution and xl list shows dying (null) domains. Over time it's causing high resource consumption and unusual test results.
Have you spotted something like that?

marmarek:
Yes, but I think pacat isn't the culprit. I've found that qubes-guid in some cases doesn't notice domain shutdown and it prevents domain cleanup. When appropriate qubes-guid is cleaned, domain is cleaned up and pacat terminates.

PR #50, opened 2020-07-23
Prevent VM notifications from appearing on top of xscreensaver

pwmarcz:
Detect a screensaver window by override_redirect, map state (an
inactive window is unmapped), and window class. Use XRestackWindows
to put newly appearing windows below such a window.

To test:
1. run in VM: while true; do notify-send "hey"; sleep 1; done
2. lock screen and check if notifications appear.

The problem doesn't seem to appear on i3lock. I haven't checked
KDE.

See QubesOS/qubes-issues#5908.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2323,6 +2323,90 @@ static void handle_wmflags(Ghandles * g, struct windowdata *vm_window)
     }
 }
 
+
+/* Check if we should keep this window on top of others */
+static bool should_keep_on_top(Ghandles *g, Window window) {
+    XWindowAttributes attr;
+    XClassHint hint;
+    bool result;
+
+    /* Check if the window has override_redirect attribute, and is mapped. */
+    XGetWindowAttributes(g->display, window, &attr);
+    if (!(attr.override_redirect && attr.map_state == IsViewable)) {
+        return false;
+    }
+
+    /* Check if this is a dom0 screensaver window by looking at window class.
+     * (VM windows have a prefix, so this is not spoofable by a VM). */
+    if (XGetClassHint(g->display, window, &hint) == 0)
+        return false;
+
+    result = (strcmp(hint.res_name, "xscreensaver") == 0);
+
+    XFree(hint.res_name);
+    XFree(hint.res_class);
+    return result;
+}
+
+
+/* Move a newly mapped override_redirect window below windows that need to be
+ * kept on top, i.e. screen lockers. */
+static void restack_windows(Ghandles *g, struct windowdata *vm_window)
+{
+    Window root;
+    Window parent;
+    Window *children_list;
+    unsigned int children_count;
+    int i, current_pos, goal_pos;
+
+    /* Find all windows below parent */
+
+    XQueryTree(
+        g->display,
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -2323,6 +2323,90 @@ static void handle_wmflags(Ghandles * g, struct windowdata *vm_window)
...
+    if (!(attr.override_redirect && attr.map_state == IsViewable)) {
+        return false;
+    }
+
+    /* Check if this is a dom0 screensaver window by looking at window class.
+     * (VM windows have a prefix, so this is not spoofable by a VM). */
+    if (XGetClassHint(g->display, window, &hint) == 0)
+        return false;
+
+    result = (strcmp(hint.res_name, "xscreensaver") == 0);
```
marmarek, line 2344: Please make this configurable (cmdline option should be enough) and allow it to be a list.

gui-daemon/xside.c
```diff
@@ -2323,6 +2323,90 @@ static void handle_wmflags(Ghandles * g, struct windowdata *vm_window)
...
+
+        if (g->log_level > 0) {
+            fprintf(stderr, "restack_windows: moving window 0x%x deeper\n",
+                    (int) vm_window->local_winid);
+        }
+
+        to_restack[0] = children_list[goal_pos];
+        to_restack[1] = vm_window->local_winid;
+
+        XRestackWindows(g->display, to_restack, 2);
```
marmarek, line 2403: There can be a race condition here - the other window may disappear in the meantime. According to the manual, this will generate `BadWindow` error and we do have `exit(1)` in the error handler. I think it should be safe to ignore this error (if window disappears, no need to restack anything), so please add appropriate exception to the error handler (`x11_error_handler()`). Extra context: [how error reporting works in xlib](https://www.x.org/releases/current/doc/libX11/libX11/libX11.html#Errors)

gui-daemon/xside.c
```diff
@@ -2323,6 +2323,90 @@ static void handle_wmflags(Ghandles * g, struct windowdata *vm_window)
...
+
+    /* Check if the window has override_redirect attribute, and is mapped. */
+    XGetWindowAttributes(g->display, window, &attr);
+    if (!(attr.override_redirect && attr.map_state == IsViewable)) {
+        return false;
+    }
+
+    /* Check if this is a dom0 screensaver window by looking at window class.
+     * (VM windows have a prefix, so this is not spoofable by a VM). */
+    if (XGetClassHint(g->display, window, &hint) == 0)
```
marmarek, line 2341: See comment about XRestackWindows error handling - I think the same applies here.

gui-daemon/xside.c
pwmarcz: Done.

gui-daemon/xside.c
pwmarcz: I modified error handler so that it ignores these (according to xlib source code, the requests are ConfigureWindow and GetProperty).

This also applies to XGetWindowAttributes above (and I was even able to crash guid on it, by spamming a lot of windows using `notify-send` in dom0). See the commit.

gui-daemon/xside.c
pwmarcz: Done, see the other comment.

PR #51, opened 2020-08-07
Update travis

fepitre:


marmarek:
Just after merging this one the travis build now fails on test-installing packages. This is because the *-dom0 packages are built for VM too, and those depends on qubes-core-dom0 that isn't in VM repository. Any idea how to solve this without reworking the whole packaging @fepitre ? Perhaps add some exclude list to travis check?

PR #53, opened 2020-08-27
icon-receiver: Retry X connection on error

rmol:
This adds an exception handler to the event loop, which will try to reestablish the xcb connection on error. The connection logic is pulled out of `IconReceiver.__init__` to a new `connect` method, to support this.

This does mitigate https://github.com/QubesOS/qubes-issues/issues/5979, but I hesitate to call it a complete fix. What happens for me is that `icon-receiver` still fails to connect for 6-7 seconds, during which time it's still spamming `.xsession-errors.old`, but it is eventually able to connect and the torrent of error logging stops.

### DIFF
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -402,11 +402,18 @@ def main():
 
     rcvd = IconReceiver()
 
+    def handle_exception(loop, context):
+        e = context.get('exception')
+        if isinstance(e, xcb.ConnectionException):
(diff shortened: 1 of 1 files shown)


marmarek:
An alternative would be to simply exit on X connection error. There is a `/etc/xdg/autostart/qubes-icon-receiver.desktop` entry that should restart it when the X server (and user session) is back.

rmol:
That does seem to work better.

marmarek (approved):

PR #56, opened 2020-12-01
More protection vs override redirect windows

DemiMarie:
This adds two features:

- A flag to disable override redirect windows entirely.  This will break
  many applications.
- A distance that any window must have from the top of the screen.  This
  prevents an override redirect window from drawing over the taskbar
  from the GUI domain.

### DIFF
--- a/gui-daemon/guid.conf
+++ b/gui-daemon/guid.conf
@@ -56,4 +56,8 @@ global: {
   # Timeout when waiting for qubes-gui-agent
   #
   # startup_timeout = 45;
+
+  # Taskbar height
+  #
+  # taskbar_height = 20
 }
--- a/gui-daemon/guid.conf
+++ b/gui-daemon/guid.conf
@@ -56,8 +56,4 @@ global: {
   # Timeout when waiting for qubes-gui-agent
   #
   # startup_timeout = 45;
-
-  # Taskbar height
-  #
-  # taskbar_height = 20
 }
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -478,9 +478,11 @@ static void update_work_area(Ghandles *g) {
     uint32_t current_desktop = (uint32_t)*scratch;
     XFree(scratch);
     ret = XGetWindowProperty(g->display, g->root_win, g->wm_workarea,
-            current_desktop * 4, 4, False, XA_CARDINAL, &act_type, &act_fmt,
+            current_desktop * desktop_coordinates_size,
+            desktop_coordinates_size, False, XA_CARDINAL, &act_type, &act_fmt,
             &nitems, &bytesleft, (unsigned char**)&scratch);
-    if (ret != Success || nitems != 4 || act_fmt != 32 || act_type != XA_CARDINAL) {
+    if (ret != Success || nitems != desktop_coordinates_size || act_fmt != 32 ||
+        act_type != XA_CARDINAL) {
         if (None == act_fmt && !act_fmt && !bytesleft) {
             g->work_x = 0;
             g->work_y = 0;
@@ -495,7 +497,7 @@ static void update_work_area(Ghandles *g) {
                 stderr);
         exit(1);
     }
-    for (int s = 0; s < 4; ++s) {
+    for (int s = 0; s < desktop_coordinates_size; ++s) {
         if (scratch[s] > max_display_width) {
             fputs("Absurd work area, crashing\n", stderr);
             abort();
@@ -1225,9 +1227,11 @@ static void moveresize_vm_window(Ghandles * g, struct windowdata *vm_window)
     if (!vm_window->is_docked) {
         /* we have window content coordinates, but XMoveResizeWindow requires
          * left top *border* pixel coordinates (if any border is present). */
-        ret = XGetWindowProperty(g->display, vm_window->local_winid, g->frame_extents, 0, 4,
-                False, XA_CARDINAL, &act_type, &act_fmt, &nitems, &bytesleft, (unsigned char**)&frame_extents);
-        if (ret == Success && nitems == 4) {
+        ret = XGetWindowProperty(g->display, vm_window->local_winid,
+                g->frame_extents, 0, desktop_coordinates_size,
+                False, XA_CARDINAL, &act_type, &act_fmt, &nitems, &bytesleft,
+                (unsigned char**)&frame_extents);
+        if (ret == Success && nitems == desktop_coordinates_size) {
             x = vm_window->x - frame_extents[0];
             y = vm_window->y - frame_extents[2];
             XFree(frame_extents);
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -3084,7 +3084,7 @@ enum {
 };
 
 struct option longopts[] = {
-    { "disable-override-redirect", no_argument, NULL, 'r' },
+    { "override-redirect", required_argument, NULL, 'r' },
     { "config", required_argument, NULL, 'C' },
     { "domid", required_argument, NULL, 'd' },
     { "name", required_argument, NULL, 'N' },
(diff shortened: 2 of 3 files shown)


marmarek (changes requested):
Some high-level comments:
 - you proposed `--override-redirect=disabled` option in https://github.com/QubesOS/qubes-issues/issues/4705#issuecomment-734500852 - IMO better to start with this option, even if only one version is implemented for now - it will avoid deprecating old option when we get more variants
 - instead of `taskbar_height` like now (which covers only taskbar at the top case), better use a generic way to avoid covering panel(s), wherever it is; I think that's [`_NET_WORKAREA` property  of the root window](https://specifications.freedesktop.org/wm-spec/1.4/ar01s03.html)

marmarek:
PipelineRetry

DemiMarie:
> * instead of `taskbar_height` like now (which covers only taskbar at the top case), better use a generic way to avoid covering panel(s), wherever it is; I think that's [`_NET_WORKAREA` property  of the root window](https://specifications.freedesktop.org/wm-spec/1.4/ar01s03.html)

How do I read this?  I can’t figure out what flags I should pass to `XGetWindowProperty`.  The documentation indicates that it is a `CARDINAL[][4]/32` but I have no idea what that refers to.

DemiMarie:
Also, can I assume `_NET_WORKAREA` to be a constant?

marmarek:
> How do I read this? I can’t figure out what flags I should pass to `XGetWindowProperty`. The documentation indicates that it is a `CARDINAL[][4]/32` but I have no idea what that refers to.

format = 32
long_length = some reasonable size for 4*number of workspaces - for example 64 (if too small, you will need to retrieve the reset with a second call)
req_type = XA_CARDINAL
property = `XinternAtom(g->display, "_NET_WORKAREA", False)`

See `process_xevent_propertynotify` function for example usage, including retrieving log value.

> Also, can I assume `_NET_WORKAREA` to be a constant?

No, user can move/resize panel(s). But you can easily get notification when it gets changed - like in `process_xevent_propertynotify`. Subscribe for properties changes on the root window with `XSelectInput()` (also already used in the code, but not on the root window).

Theoretically this also means, you'll need `_NET_CURRENT_DESKTOP`. But IMO it is quite safe to assume every workspace have the same areas (which you can also verify) and you can simply use the first one.

DemiMarie:
So my current plan is to use `_NET_CURRENT_DESKTOP` to determine the correct offset in `_NET_WORKAREA` to use.  Since a user could have multiple monitors, the workspaces cannot (IMO) be assumed to have the same area.  Fortunately, out of bounds offsets are just an error, not undefined behavior, so I am not too worried there.

How important is watching for property changes rather than re-fetching property values?

marmarek:
IIUC this gets called only for override-redirect windows (which rarely move), right? In that case, I think re-fetching may be ok.

DemiMarie:
Correct.  I can’t test this code locally at all, and am no expert on either the GUI daemon or X11.  Therefore, I strongly desire to keep the code as simple and obviously correct as possible, to minimize the chance that a bug will get through.

If possible, I would prefer that this code be tested before it goes live.  I can try to test it on the test machine, but it will take me a while to set it up.

gui-daemon/xside.c
```diff
@@ -434,6 +434,73 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
...
+        /* Panic!  Serious window manager problem. */
+        fputs("PANIC: cannot obtain current desktop\n"
+                "Instead of creating a security hole we will just exit.\n",
+            stderr);
+        exit(1);
+    }
+    uint32_t current_desktop = (uint32_t)*scratch;
+    XFree(scratch);
+    ret = XGetWindowProperty(g->display, g->root_win, g->wm_workarea,
+            current_desktop, 4, False, XA_CARDINAL, &act_type, &act_fmt,
```
marmarek, line 474:
```suggestion
            current_desktop * 4, 4, False, XA_CARDINAL, &act_type, &act_fmt,
```

gui-daemon/guid.conf
```diff
@@ -56,4 +56,8 @@ global: {
   # Timeout when waiting for qubes-gui-agent
   #
   # startup_timeout = 45;
+
+  # Taskbar height
+  #
+  # taskbar_height = 20
```
marmarek, line 62: This also needs an update.

marmarek (changes requested):

marmarek:
CI:
```
DEBUG: xside.c:447:18: error: 'desktop_coordinates_size' defined but not used [-Werror=unused-const-variable=]
DEBUG:   447 | static const int desktop_coordinates_size = 4;
DEBUG:       |                  ^~~~~~~~~~~~~~~~~~~~~~~~
DEBUG: cc1: all warnings being treated as errors
```

gui-daemon/xside.c
```diff
@@ -454,28 +528,29 @@ static void mkghandles(Ghandles * g)
...
-    g->wm_state_maximized_horz = XInternAtom(g->display, "_NET_WM_STATE_MAXIMIZED_HORZ", False);
+    g->tray_selection = intern_atom(g, tray_sel_atom_name, False);
+    g->tray_opcode = intern_atom(g, "_NET_SYSTEM_TRAY_OPCODE", False);
+    g->xembed_message = intern_atom(g, "_XEMBED", False);
+    g->xembed_info = intern_atom(g, "_XEMBED_INFO", False);
+    g->wm_state = intern_atom(g, "_NET_WM_STATE", False);
+    g->wm_state_fullscreen = intern_atom(g, "_NET_WM_STATE_FULLSCREEN", False);
+    g->wm_state_demands_attention = intern_atom(g, "_NET_WM_STATE_DEMANDS_ATTENTION", False);
+    g->wm_state_hidden = intern_atom(g, "_NET_WM_STATE_HIDDEN", False);
+    g->wm_workarea = intern_atom(g, "_NET_WORKAREA", True);
```
marmarek, line 544: `True` here makes it fail hard if `_NET_WORKAREA` is not supported by the WM.

And BTW we do have such case - with hybrid GUI domain, where gui-daemon in dom0 for sys-gui is launched without window manager at all, and from dom0's Xorg point of view, there is just one fullscreen window.

marmarek (changes requested):

gui-daemon/xside.c
```diff
@@ -716,8 +794,10 @@ static int run_clipboard_rpc(Ghandles * g, enum clipboard_op op) {
             }
             dup2(fd, 0);
             close(fd);
-            snprintf(domid_str, sizeof(domid_str), "%d", g->target_domid);
-            execl(QREXEC_CLIENT_PATH, "qrexec-client", "-d", domid_str, service_call, (char*)NULL);
+            if ((unsigned)snprintf(domid_str, sizeof(domid_str), "%d", g->target_domid)
+                >= sizeof(domid_str))
+                abort();
+            execl(QREXEC_CLIENT_PATH, "qrexec-client", "-tT", "-d", domid_str, service_call, (char*)NULL);
```
marmarek, line 800: `-t` here breaks unicode characters in clipboard data.

DemiMarie: oops :laughing:, fixed by only using `-T`.  Also, this is broken with a GUI domain outside of dom0.

marmarek: In fact, clipboard looks to be broken with GUI domain even when using standard method, not qrexec. Because gui-daemon still tries to consult qrexec policy.

DemiMarie: And it must consult qrexec policy ― that’s why `qubes.ClipboardPaste` exists.  Hard-coding it to `@anyvm @anyvm ask` is not a good idea.

marmarek (approved):

gui-daemon/xside.c
```diff
@@ -1187,41 +1267,42 @@ static int force_on_screen(Ghandles * g, struct windowdata *vm_window,
     int x = vm_window->x, y = vm_window->y, w = vm_window->width, h =
         vm_window->height;
 
-    if (vm_window->x < border_width
-        && vm_window->x + (int)vm_window->width > 0) {
-        vm_window->x = border_width;
+    const int border_x = border_width + g->work_x,
+        border_y = border_width + g->work_y,
```
iamahuman, line 1271: This should be `border_height`

PR #57, opened 2020-12-09
Fix clipboard stop working after ~50 days of uptime

marmarek:
We use X11 event time to detect if any other clipboard operation
happened between user trigger (via key combo) and the actual operation
happens (like - VM sends the clipboard content). But when the X11 timer
wraps (every 2^32 ms = ~50 days), the check fails as the "clipboard
file" event time (just before the wrap) will be greater than the current
event time (just after the wrap).
Detect the wrap by checking if the difference is no larger than half of
the timer range.

Theoretically this could still fail if a single clipboard operation will
take ~25 days _and no other clipboard operation happens in the meantime_.
If another operation happens in the meantime, it _should_ fail.

Fixes QubesOS/qubes-issues#2189

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -799,7 +799,9 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     }
     inter_appviewer_lock(g, 1);
     clipboard_file_xevent_time = get_clipboard_file_xevent_timestamp();
-    if (clipboard_file_xevent_time > g->clipboard_xevent_time) {
+    /* X11 time is just 32-bit miliseconds counter, which make it wrap every
+     * ~50 days - something that is realistic. Handle that wrapping too. */
+    if (clipboard_file_xevent_time - g->clipboard_xevent_time < (1UL<<31)) {
         /* some other clipboard operation happened in the meantime, discard
          * request */
         inter_appviewer_lock(g, 0);
@@ -944,7 +946,9 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
             return 1;
         inter_appviewer_lock(g, 1);
         clipboard_file_xevent_time = get_clipboard_file_xevent_timestamp();
-        if (clipboard_file_xevent_time > ev->time) {
+        /* X11 time is just 32-bit miliseconds counter, which make it wrap every
+         * ~50 days - something that is realistic. Handle that wrapping too. */
+        if (clipboard_file_xevent_time - ev->time < (1UL<<31)) {
             /* some other clipboard operation happened in the meantime, discard
              * request */
             inter_appviewer_lock(g, 0);


marmarek:
@DemiMarie can you review (apparently github allows to request reviews only from users with write access...)

gui-daemon/xside.c
```diff
@@ -944,7 +947,10 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
             return 1;
         inter_appviewer_lock(g, 1);
         clipboard_file_xevent_time = get_clipboard_file_xevent_timestamp();
-        if (clipboard_file_xevent_time > ev->time) {
+        /* X11 time is just 32-bit miliseconds counter, which make it wrap every
```
DemiMarie, line 950: What happens if `clipboard_file_xevent_time` wraps before `ev->time` does?  In that case, the check could pass when it should not.  Additionally, this code appears to be duplicated below, and we might want to replace timestamps with a mutex or other lock.

DemiMarie (changes requested):
If `clipboard_file_xevent_time` is the first to wrap, subsequent clipboard changes will not be detected.

gui-daemon/xside.c
marmarek: > What happens if `clipboard_file_xevent_time` wraps before `ev->time` does?

There is very narrow window for this to happen: user-triggered paste (`ev->time`) needs to happen just before the wrap and then just after wrap some other clipboard operation would need to be initiated (by the user) and completed, all this before `inter_appviewer_lock()` call, so basically no blocking calls for this time. While theoretically possible, technically not (user racing against themself).

This issue is more likely to happen in the other place - where indeed the time between those two can be longer. I think we should require not only ">" relation but also the difference being not greater than 2^31 or similar.
Since those are unsigned values, I think it would be enough to just: `clipboard_file_xevent_time - ev->time < (1UL<<31)`, correct?

> and we might want to replace timestamps with a mutex or other lock.

While it would be immune to the time wrap issue, it would introduce new vulnerability: a VM could prevent other VMs to perform clipboard operations (inter-vm clipboard is a global resource). This is at least denial-of-service, but likely could be used to delay an operation in another VM by some specific time to cause some harm.

DemiMarie: > > What happens if `clipboard_file_xevent_time` wraps before `ev->time` does?
> 
> There is very narrow window for this to happen: user-triggered paste (`ev->time`) needs to happen just before the wrap and then just after wrap some other clipboard operation would need to be initiated (by the user) and completed, all this before `inter_appviewer_lock()` call, so basically no blocking calls for this time. While theoretically possible, technically not (user racing against themself).
> 
> This issue is more likely to happen in the other place - where indeed the time between those two can be longer. I think we should require not only ">" relation but also the difference being not greater than 2^31 or similar.
> Since those are unsigned values, I think it would be enough to just: `clipboard_file_xevent_time - ev->time < (1UL<<31)`, correct?

I believe so, but we might want to use a version that is more obviously correct.
 
> > and we might want to replace timestamps with a mutex or other lock.
> 
> While it would be immune to the time wrap issue, it would introduce new vulnerability: a VM could prevent other VMs to perform clipboard operations (inter-vm clipboard is a global resource). This is at least denial-of-service, but likely could be used to delay an operation in another VM by some specific time to cause some harm.

When copying, we can avoid this problem by using very short timeouts, since a legitimate VM will respond extremely quickly.  On error, we can fail-safe by copying the empty string.  When pasting, we can read the entire clipboard contents into memory in dom0; the VM can then grab them at will.

marmarek: I don't like the timeout approach, because it is very hard to get the arbitrary timeout value right. If too short - some busy VM may fail to respond in time (leading to copy/paste fail) - for example a VM that is heavily swapping (happens with Firefox more often recently...), I hit cases where it took up to few seconds. But with timeout too long, one VM may block another for some time (which will be an UX issue even without malicious context).

What we need here is to verify the actions happened in specific order, not that they were fast.

marmarek:
PipelineRetry

gui-daemon/xside.c
DemiMarie: `clipboard_file_xevent_time - ev->time < (1UL<<31)` is fine.

DemiMarie (approved):

PR #59, opened 2021-05-04
Call policy.EvalSimple when outside of dom0

DemiMarie:
Outside of dom0, `qrexec-policy` is not available, but we can use
`qrexec-client-vm` and the `policy.EvalGUI` service instead.

Fixes QubesOS/qubes-issues#6263.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -28,9 +28,13 @@
 #include <err.h>
 #include <sys/shm.h>
 #include <sys/file.h>
+#include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/socket.h>
+#include <sys/un.h>
 #include <sys/wait.h>
 #include <sys/resource.h>
+#include <sys/uio.h>
 #include <signal.h>
 #include <poll.h>
 #include <errno.h>
@@ -748,12 +752,14 @@ static int run_clipboard_rpc(Ghandles * g, enum clipboard_op op) {
         case CLIPBOARD_COPY:
             path_stdin = "/dev/null";
             path_stdout = QUBES_CLIPBOARD_FILENAME;
-            service_call = "DEFAULT:QUBESRPC qubes.ClipboardCopy";
+            service_call = g->in_dom0 ? QREXEC_COMMAND_PREFIX QUBES_SERVICE_CLIPBOARD_COPY :
+                QUBES_SERVICE_CLIPBOARD_COPY;
             break;
         case CLIPBOARD_PASTE:
             path_stdin = QUBES_CLIPBOARD_FILENAME;
             path_stdout = "/dev/null";
-            service_call = "DEFAULT:QUBESRPC qubes.ClipboardPaste";
+            service_call = g->in_dom0 ? QREXEC_COMMAND_PREFIX QUBES_SERVICE_CLIPBOARD_PASTE :
+                QUBES_SERVICE_CLIPBOARD_PASTE;
             break;
         default:
             /* not reachable */
@@ -770,7 +776,7 @@ static int run_clipboard_rpc(Ghandles * g, enum clipboard_op op) {
 
             /* grant group write */
             old_umask = umask(0007);
-            fd = open(path_stdout, O_WRONLY|O_CREAT|O_TRUNC, 0644);
+            fd = open(path_stdout, O_WRONLY|O_CREAT|O_TRUNC|O_NOCTTY, 0644);
             if (fd < 0) {
                 perror("open");
                 _exit(1);
@@ -783,23 +789,26 @@ static int run_clipboard_rpc(Ghandles * g, enum clipboard_op op) {
             }
             dup2(fd, 1);
             close(fd);
-            fd = open(path_stdin, O_RDONLY);
+            fd = open(path_stdin, O_RDONLY|O_NOCTTY);
             if (fd < 0) {
                 perror("open");
                 _exit(1);
             }
             dup2(fd, 0);
             close(fd);
-            if ((unsigned)snprintf(domid_str, sizeof(domid_str), "%d", g->target_domid)
-                >= sizeof(domid_str))
-                abort();
-            execl(QREXEC_CLIENT_PATH, "qrexec-client", "-T", "-d", domid_str, service_call, (char*)NULL);
+            if (g->in_dom0) {
+                if ((unsigned)snprintf(domid_str, sizeof(domid_str), "%d", g->target_domid)
+                    >= sizeof(domid_str))
+                    abort();
+                execl(QREXEC_CLIENT_PATH, QREXEC_CLIENT, "-T", "-d", domid_str, service_call, (char*)NULL);
+            } else
+                execl(QREXEC_CLIENT_VM_PATH, QREXEC_CLIENT_VM, "-T", "--", g->vmname, service_call, (char*)NULL);
             perror("execl");
             _exit(1);
         default:
             waitpid(pid, &status, 0);
     }
-    return WEXITSTATUS(status) == 0;
+    return WIFEXITED(status) && WEXITSTATUS(status) == 0;
 }
 
 static int fetch_qubes_clipboard_using_qrexec(Ghandles * g) {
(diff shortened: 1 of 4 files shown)


marmarek:
This fails to build - see CI

DemiMarie:
This should now work in all cases, when combined with QubesOS/qubes-core-qrexec#69.  I changed the code so that it just connects to the `policy.Eval` socket if it is in dom0.

marmarek:
Tests:
```
May 23 00:58:38 dom0 qrexec-policy-daemon[1252]: b'policy.EvalSimple': invalid data from qube
```

DemiMarie:
> Tests:
> 
> ```
> May 23 00:58:38 dom0 qrexec-policy-daemon[1252]: b'policy.EvalSimple': invalid data from qube
> ```

I will try to reproduce this.

marmarek:
The dom0 case works now!

But the guivm doesn't (https://openqa.qubes-os.org/tests/17944). `guid.personal.log` from inside sys-gui:
```
dup2: Resource temporarily unavailable
recv: Connection reset by peer
```
And I don't see it made to the qrexec-policy-daemon.

DemiMarie:
> The dom0 case works now!
> 
> But the guivm doesn't (https://openqa.qubes-os.org/tests/17944). `guid.personal.log` from inside sys-gui:
> 
> ```
> dup2: Resource temporarily unavailable
> recv: Connection reset by peer
> ```
> 
> And I don't see it made to the qrexec-policy-daemon.

:woman_facepalming: `dup2()` returns the new file descriptor on success, not 0.  Easy fix.

marmarek:
GUIVM case:
```
qrexec-client-vm: unrecognized option '--@adminvm'
usage: qrexec-client-vm [options] target_vmname program_ident [local_program [local program arguments]]
Options:
  --buffer-size=BUFFER_SIZE - minimum vchan buffer size (default: 64k)
  -t, --filter-escape-chars-stdout - filter non-ASCII and control characters on stdout (default if stdout is a terminal)
  -T, --filter-escape-chars-stderr - filter non-ASCII and control characters on stderr (default if stderr is a terminal)
  --no-filter-escape-chars-stdout - opposite to --filter-escape-chars-stdout
  --no-filter-escape-chars-stderr - opposite to --filter-escape-chars-stderr
  --agent-socket=PATH - path to connect to, default: /var/run/qubes/qrexec-agent
recv: Connection reset by peer
```

But also, the gui daemon was terminated. Lets to it this way: for each `exit()` you add, accompany it with a comment why it is absolutely impossible or undesirable to recover from this situation - otherwise remove the `exit()` and add graceful recovery.

DemiMarie:
> GUIVM case:
> 
> ```
> qrexec-client-vm: unrecognized option '--@adminvm'
> usage: qrexec-client-vm [options] target_vmname program_ident [local_program [local program arguments]]
> Options:
>   --buffer-size=BUFFER_SIZE - minimum vchan buffer size (default: 64k)
>   -t, --filter-escape-chars-stdout - filter non-ASCII and control characters on stdout (default if stdout is a terminal)
>   -T, --filter-escape-chars-stderr - filter non-ASCII and control characters on stderr (default if stderr is a terminal)
>   --no-filter-escape-chars-stdout - opposite to --filter-escape-chars-stdout
>   --no-filter-escape-chars-stderr - opposite to --filter-escape-chars-stderr
>   --agent-socket=PATH - path to connect to, default: /var/run/qubes/qrexec-agent
> recv: Connection reset by peer
> ```

Fixed (missing comma)

> But also, the gui daemon was terminated. Lets to it this way: for each `exit()` you add, accompany it with a comment why it is absolutely impossible or undesirable to recover from this situation - otherwise remove the `exit()` and add graceful recovery.

That is a good approach!

DemiMarie:
PipelineRetry

marmarek:
```
Operation refused by dom0
qrexec-client-vm failed
```
But no info from the policy daemon.
I guess it's time for a proper review, instead of a poor approximation of test driven development ;)

DemiMarie:
> ```
> Operation refused by dom0
> qrexec-client-vm failed
> ```
> 
> But no info from the policy daemon.
> I guess it's time for a proper review, instead of a poor approximation of test driven development ;)

Sorry about this.  I would like to hope that my code is normally better than this 🙂.

gui-daemon/xside.c
```diff
@@ -910,11 +919,146 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+        case -1:
+            perror("fork");
+            exit(1); // not worth handling can’t fork
+        case 0:
+            if (dup2(sockets[0], 0) == -1 || dup2(sockets[0], 1) == -1) {
+                perror("dup2");
+                _exit(1);
+            }
+            execl(QREXEC_CLIENT_VM_PATH, QREXEC_CLIENT_VM, "--",
+                    "@adminvm", "policy.EvalGUI+ClipboardPaste", NULL);
```
marmarek, line 941: This one still doesn't use the macro (and is wrong).

DemiMarie: Oops!  That is the last one, thankfully.

gui-daemon/xside.c
```diff
@@ -910,11 +919,147 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+        Ghandles *g, int socket, const char *const source_vm) {
+    char buf[63];
+    size_t const source_vm_len = strlen(source_vm);
+    size_t const dst_vm_len = strlen(g->vmname);
+    if (source_vm_len > 31 || dst_vm_len > 31) {
+        fputs("VM name too long\n", stderr);
+        exit(1); // this means something has gone horribly wrong elsewhere
+    }
+    memcpy(buf, source_vm, source_vm_len + 1);
+    memcpy(buf + source_vm_len + 1, g->vmname, dst_vm_len);
```
marmarek, line 1020: I think you got this right, but I'd still prefer `sprintf`, for readability (you'll need to make `buf` 1 byte longer for that). And you can send one byte less below, to avoid sending extra NUL char.

gui-daemon/xside.c
```diff
@@ -910,11 +919,147 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+    if (!write_all(socket, buf, source_vm_len + dst_vm_len + 1))
+        return false;
+    if (shutdown(socket, SHUT_WR)) {
+        perror("shutdown");
+        abort(); // cannot happen
```
marmarek, line 1025: This is really not a reason for crashing the whole process... Simply close the FD and treat as policy denial (after reporting the error).

gui-daemon/xside.c
```diff
@@ -910,11 +919,147 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+        if (recv_len == 0)
+            break;
+        ssize_t res = recv(socket, buf + bytes_received, recv_len, MSG_WAITALL);
+        if (!res)
+            break; /* EOF */
+        if (res == -1) {
+            if (errno == EINTR)
+                continue;
+            perror("recv");
+            abort(); /* cannot happen per man page */
```
marmarek, line 1040: Again, there is no need to crash here. In fact, this can (and do) happen - various socket families can define additional error codes - and I hit `ECONNRESET` here already.

gui-daemon/xside.c
```diff
@@ -910,11 +919,147 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+        }
+        if ((size_t)res > recv_len)
+            abort();
+        bytes_received += (size_t)res;
+    }
+    if (bytes_received == QUBES_POLICY_ACCESS_ALLOWED_LEN &&
+             !memcmp(buf, QUBES_POLICY_ACCESS_ALLOWED, QUBES_POLICY_ACCESS_ALLOWED_LEN))
+        return true;
```
marmarek, line 1048: For future extensibility, can we compare a prefix (including `\n`) only?

marmarek (changes requested):

gui-daemon/xside.c
DemiMarie: Sure.

gui-daemon/xside.c
DemiMarie: Fixed

marmarek (approved):

PR #61, opened 2021-05-25
Move XInternAtom calls to startup

DemiMarie:
This should make some operations a little bit faster.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -491,11 +491,56 @@ static void update_work_area(Ghandles *g) {
     XFree(scratch);
 }
 
+/*
+ * Internal all of the atoms we will use.  For performance reasons, we perform
+ * all atom interning at startup, and do so using a single XInternAtoms() call.
+ */
+static void intern_global_atoms(Ghandles *const g) {
+    char tray_sel_atom_name[64];
+    if ((unsigned)snprintf(tray_sel_atom_name, sizeof(tray_sel_atom_name),
+        "_NET_SYSTEM_TRAY_S%u", DefaultScreen(g->display)) >=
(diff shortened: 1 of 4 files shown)


gui-daemon/xside.c
```diff
@@ -526,26 +511,48 @@ static void mkghandles(Ghandles * g)
...
+            { &g->qubes_vmwindowid, "_QUBES_VMWINDOWID" },
+            { &g->net_wm_icon, "_NET_WM_ICON" },
+            { &g->wm_current_desktop, "_NET_WM_CURRENT_DESKTOP" },
+            { &g->wmDeleteMessage, "WM_DELETE_WINDOW" },
+        };
+        Atom labels[QUBES_ARRAY_SIZE(atoms_to_intern)];
+        const char *names[QUBES_ARRAY_SIZE(atoms_to_intern)];
+        for (size_t i = 0; i < QUBES_ARRAY_SIZE(atoms_to_intern); ++i)
+            names[i] = atoms_to_intern[i].name;
+        if (!XInternAtoms(g->display, (char **)names, QUBES_ARRAY_SIZE(atoms_to_intern), 1, labels)) {
```
marmarek, line 549: `only_if_exists` param should be `False`, not `1`. Have you tried to run this?

marmarek (changes requested):

gui-daemon/xside.c
DemiMarie: No, but I probably should before asking you to review it :woman_facepalming:.

DemiMarie:
Tested (and working) on the test laptop.

marmarek (approved):

PR #63, opened 2021-06-05
Avoid rogue screen locking

DemiMarie:
by setting the _NET_WM_USER_TIME property when needed.

Fixes QubesOS/qubes-issues#3727.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -358,7 +358,7 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
                     vm_window->x, vm_window->y,
                     vm_window->width,
                     vm_window->height, 0,
-		    CopyFromParent,
+                    CopyFromParent,
                     CopyFromParent,
                     CopyFromParent,
                     CWOverrideRedirect, &attr);
@@ -367,6 +367,11 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
     (void) XSetStandardProperties(g->display, child_win,
                       "VMapp command", "Pixmap", None,
                       gargv, 0, &my_size_hints);
+    if (g->time_win != None)
+        XChangeProperty(g->display, child_win, g->wm_user_time_window,
+               XA_WINDOW, 32, PropModeReplace,
+               (const unsigned char *)&g->time_win,
+               1);
     (void) XSelectInput(g->display, child_win,
                 ExposureMask | KeyPressMask | KeyReleaseMask |
                 ButtonPressMask | ButtonReleaseMask |
@@ -527,7 +532,10 @@ static void intern_global_atoms(Ghandles *const g) {
         { &g->qubes_vmwindowid, "_QUBES_VMWINDOWID" },
         { &g->net_wm_icon, "_NET_WM_ICON" },
         { &g->wm_current_desktop, "_NET_WM_CURRENT_DESKTOP" },
+        { &g->wm_user_time_window, "_NET_WM_USER_TIME_WINDOW" },
+        { &g->wm_user_time, "_NET_WM_USER_TIME" },
         { &g->wmDeleteMessage, "WM_DELETE_WINDOW" },
+        { &g->net_supported, "_NET_SUPPORTED" },
     };
     Atom labels[QUBES_ARRAY_SIZE(atoms_to_intern)];
     const char *names[QUBES_ARRAY_SIZE(atoms_to_intern)];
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -600,11 +641,45 @@ static void mkghandles(Ghandles * g)
...
+            if (XGetWindowProperty(g->display, g->root_win, g->wm_user_time_window,
+                    0, 1, False, XA_WINDOW, &act_type, &act_fmt,
+                    &nitems, &bytesleft, (unsigned char**)&scratch) != Success ||
+                act_type != XA_WINDOW ||
+                act_fmt != 32 ||
+                bytesleft != 0 ||
+                nitems != 1) {
+                if (act_type == None && act_fmt == 0 && bytesleft == 0) {
+                    /* Not supported? */
+                    fputs("_NET_WM_USER_TIME_WINDOW supported, but not found on the root window?\n",
```
marmarek, line 670: I think you got the spec wrong. It should be set on application's window(s), not on the root window.
See https://specifications.freedesktop.org/wm-spec/1.4/ar01s05.html

Spec gives two options:
 - setting `_NET_WM_USER_TIME` directly on the application window (then update_wm_user_time should operate on the window where the event arrived)
 - setting `_NET_WM_USER_TIME` on a separate window, pointed with `_NET_WM_USER_TIME_WINDOW` (if window manager supports it) - in this case, gui-daemon would need to create separate stub window to set this property on (that window doesn't need to be mapped), I think it should be fine to use one stub window per gui-daemon and let all windows of that gui-daemon to point at it with `_NET_WM_USER_TIME_WINDOW`.

gui-daemon/xside.c
```diff
@@ -1541,6 +1624,7 @@ static void process_xevent_motion(Ghandles * g, const XMotionEvent * ev)
     struct msg_hdr hdr;
     struct msg_motion k;
     CHECK_NONMANAGED_WINDOW(g, ev->window);
+    update_wm_user_time(g, ev->time);
```
marmarek, line 1627: I'm not sure about this one. I think this is allowed according to the spec, but this will spam with PropertyChange events a lot. And indeed some random choice of applications (xfce4 terminal, gnome terminal, firefox) do not update the user time property on mouse movement, only on key presses + clicks.

marmarek (changes requested):

gui-daemon/xside.c
DemiMarie: That makes more sense!  I misread the specification and thought that setting `_NET_WM_USER_TIME_WINDOW` was the window manager’s job.

gui-daemon/xside.c
DemiMarie: The purpose of this is to detect scroll events, which otherwise are not detected.

marmarek: Scroll is in fact a mouse button press in X11 (buttons 4 and 5 for up/down).

marmarek:
You need to set `_NET_WM_USER_TIME_WINDOW` on app windows, pointing at `g->time_win` (if supported and created).

marmarek (approved):
Ok, now it seems correct. It is time for testing it now. AFAIR the proper test scenario for this is:
1. open some VM app
2. give it a focus, place mouse inside its window
3. do only clicks, not movements, not keyboard input
4. screen should _not_ lock after defined time (10 minutes by default)

Is that correct? Should the point 3 be about any clicks, or just scroll?

DemiMarie:
> Ok, now it seems correct. It is time for testing it now. AFAIR the proper test scenario for this is:
> 
> 1. open some VM app
> 2. give it a focus, place mouse inside its window
> 3. do only clicks, not movements, not keyboard input
> 4. screen should _not_ lock after defined time (10 minutes by default)
> 
> Is that correct? Should the point 3 be about any clicks, or just scroll?

I imagine so.  Can this be part of the integration tests on OpenQA?

marmarek:
> I imagine so. Can this be part of the integration tests on OpenQA?

That's exactly my intention.

gui-daemon/xside.c
```diff
@@ -527,7 +532,11 @@ static void intern_global_atoms(Ghandles *const g) {
         { &g->qubes_vmwindowid, "_QUBES_VMWINDOWID" },
         { &g->net_wm_icon, "_NET_WM_ICON" },
         { &g->wm_current_desktop, "_NET_WM_CURRENT_DESKTOP" },
+        { &g->wm_current_desktop, "_NET_WM_CURRENT_DESKTOP" },
```
marmarek, line 535: Rebase went wrong here.

gui-daemon/xside.c
```diff
@@ -541,6 +549,32 @@ static void intern_global_atoms(Ghandles *const g) {
...
+        Window const window, Atom property, long **state_list, size_t *items) {
+    assert(state_list && "NULL state_list in qubes_get_all_atom_properties");
+    assert(items && "NULL items in qubes_get_all_atom_properties");
+    *items = 0;
+    unsigned long nitems = 0, bytesleft = 0;
+    Atom act_type = None;
+    int act_fmt = 0, ret;
+    do {
+        /* Ensure we read all of the atoms */
+        XFree(state_list);
```
marmarek, line 562: This crashes on the first iteration.

DemiMarie: Fixed, but not pushed yet, as I ran into a `BadValue` error.

marmarek:
> That's exactly my intention.

I've implemented it at https://github.com/marmarek/openqa-tests-qubesos/commit/e5ed6c5295fdb76e100f280e6e7124f94ea0f9f9. Without this PR, it fails as expected. I've tried to test with this PR, but ran into the issue above.

DemiMarie:
I think I have a fix, but please do not review the code yet, as I forgot to test it first.

gui-daemon/xside.c
DemiMarie: Tested and working

marmarek (approved):

PR #64, opened 2021-06-24
Better build system

DemiMarie:
This significantly improves the build system:

- Header dependencies are properly tracked
- Fewer calls to pkg-config
- Less code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@
 
 MANDIR ?= /usr/share/man
 LIBDIR ?= /usr/lib64
-export LIBDIR
+VCHAN_PKG ?= vchan-xen
+export LIBDIR CFLAGS
 
 help:
 	@echo "Qubes GUI main Makefile:" ;\
@@ -30,11 +31,12 @@ help:
 	    echo "make clean                <--- clean all the binary files";\
 	    exit 0;
 
-all: gui-daemon/qubes-guid shmoverride/shmoverride.so shmoverride/X-wrapper-qubes \
-		pulse/pacat-simple-vchan screen-layout-handler/watch-screen-layout-changes
+all: gui-daemon/qubes-guid gui-daemon/qubes-guid.1 shmoverride/shmoverride.so \
+     shmoverride/X-wrapper-qubes pulse/pacat-simple-vchan \
+     screen-layout-handler/watch-screen-layout-changes
 
-gui-daemon/qubes-guid:
-	(cd gui-daemon; $(MAKE))
+gui-daemon/qubes-guid gui-daemon/qubes-guid.1:
+	$(MAKE) -C gui-daemon qubes-guid qubes-guid.1
 
 shmoverride/shmoverride.so:
 	(cd shmoverride; $(MAKE) shmoverride.so)
--- a/gui-daemon/Makefile
+++ b/gui-daemon/Makefile
@@ -19,27 +19,26 @@
 #
 #
 
+MAKEFLAGS := -rR
 VCHAN_PKG = $(if $(BACKEND_VMM),vchan-$(BACKEND_VMM),vchan)
 CC=gcc
-CFLAGS+=-I../include/ -g -O2 -Wall -Wextra -Werror -pie -fPIC \
-		`pkg-config --cflags libconfig` \
-		`pkg-config --cflags libnotify` \
-		`pkg-config --cflags libpng` \
-		`pkg-config --cflags $(VCHAN_PKG)`
-OBJS=xside.o png.o trayicon.o ../gui-common/double-buffer.o ../gui-common/txrx-vchan.o \
-	../gui-common/error.o ../common/list.o
-
-all: qubes-guid qubes-guid.1
-qubes-guid: $(OBJS)
-	$(CC) -g -pie -o qubes-guid $(OBJS) \
-		-lX11 -lXext -Wall -lm \
-		`pkg-config --libs $(VCHAN_PKG)` \
-		`pkg-config --libs libpng` \
-		`pkg-config --libs libconfig` \
-		`pkg-config --libs libnotify`
+pkgs := x11 xext glib-2.0 $(VCHAN_PKG) libpng libnotify libconfig
+objs := xside.o png.o trayicon.o ../gui-common/double-buffer.o ../gui-common/txrx-vchan.o \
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -30,11 +31,14 @@ help:
...
-all: gui-daemon/qubes-guid shmoverride/shmoverride.so shmoverride/X-wrapper-qubes \
-		pulse/pacat-simple-vchan screen-layout-handler/watch-screen-layout-changes
+all: gui-daemon/qubes-guid gui-daemon/qubes-guid.1 shmoverride/shmoverride.so \
+     shmoverride/X-wrapper-qubes pulse/pacat-simple-vchan \
+     screen-layout-handler/watch-screen-layout-changes
 
-gui-daemon/qubes-guid:
-	(cd gui-daemon; $(MAKE))
+gui-daemon/qubes-guid gui-daemon/qubes-guid.1:
+	$(MAKE) -C gui-daemon 'CFLAGS=$(subst ','\'',$(CFLAGS)) \
```
marmarek, line 39: Why set it this way, instead of proper assignment in `gui-daemon/Makefile`? The need for this subst smells like something went wrong.

Makefile
```diff
@@ -30,11 +31,14 @@ help:
...
+	-I../../gui-common/include -I../../core-qubesdb/include' \
```
marmarek, line 40: This is absolutely wrong this to do. Those should be used from appropriate -devel package installed as part of build dependencies. Even as a fallback, if coincidentally those paths are valid, the resulting binaries may be surprising - like using includes from R4.1 while building for R4.0...

If you want to build without qubes-builder (or other tool preparing isolated build env), you can add such flags using some extra variable (see comment about EXTRA_CFLAGS), empty by default. But please keep the default build clear of using undefined build dependencies.

And BTW, if you want to influence CFLAGS from in manual make call, without breaking stuff, set it as env variable, like this
```
CFLAGS="my-awesome-flags" make
```
instead of a make argument.

gui-daemon/Makefile
```diff
@@ -19,27 +19,25 @@
...
 VCHAN_PKG = $(if $(BACKEND_VMM),vchan-$(BACKEND_VMM),vchan)
 CC=gcc
-CFLAGS+=-I../include/ -g -O2 -Wall -Wextra -Werror -pie -fPIC \
-		`pkg-config --cflags libconfig` \
-		`pkg-config --cflags libnotify` \
-		`pkg-config --cflags libpng` \
-		`pkg-config --cflags $(VCHAN_PKG)`
-OBJS=xside.o png.o trayicon.o ../gui-common/double-buffer.o ../gui-common/txrx-vchan.o \
+override PKGS := x11 xext glib-2.0 $(VCHAN_PKG) libpng libnotify libconfig
+override OBJS := xside.o png.o trayicon.o ../gui-common/double-buffer.o ../gui-common/txrx-vchan.o \
```
marmarek, line 26: Why `override` for those two? Using `override` keyword in most cases means you are doing something wrong - like working around variable name clash. But I think in this very case, you can simply drop the keyword.

gui-daemon/Makefile
```diff
@@ -19,27 +19,25 @@
...
 	../gui-common/error.o ../common/list.o
-
-all: qubes-guid qubes-guid.1
+override CFLAGS := $(CFLAGS) -I../include/ -g -O2 -Wall -Wextra -Werror -pie -fPIC \
```
marmarek, line 28: Here too, if you drop setting `CFLAGS` via a command line in the main `Makefile`, the `override` here is not needed.
And BTW, better way to deal with adding such extra entries to CFLAGS, is using another variable - like setting EXTRA_CFLAGS on the outside and including it in CFLAGS here (possibly in for some targets only if necessary).

shmoverride/Makefile
```diff
@@ -27,9 +27,10 @@ $(error shmoverride currently supports Xen only, '$(BACKEND_VMM)' set)
 endif
 
 LIBDIR ?= /usr/lib64
-CFLAGS+=-g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
+override CFLAGS+=-g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
```
marmarek, line 30: same comment about override.

marmarek (changes requested):

shmoverride/Makefile
```diff
@@ -27,9 +27,10 @@ $(error shmoverride currently supports Xen only, '$(BACKEND_VMM)' set)
 endif
 
 LIBDIR ?= /usr/lib64
-CFLAGS+=-g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
+extra_cflags := -g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
 		-DBACKEND_VMM_$(BACKEND_VMM) \
-		-DSHMOVERRIDE_LIB_PATH=\"$(LIBDIR)/qubes-gui-daemon/shmoverride.so\"
+		-DSHMOVERRIDE_LIB_PATH=\"$(LIBDIR)/qubes-gui-daemon/shmoverride.so\" \
+		-I../../gui-common/include
```
marmarek, line 33: Remove this one too.

marmarek (changes requested):

marmarek (approved):

shmoverride/Makefile
```diff
@@ -27,7 +27,7 @@ $(error shmoverride currently supports Xen only, '$(BACKEND_VMM)' set)
 endif
 
 LIBDIR ?= /usr/lib64
-CFLAGS+=-g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
+extra_cflags := -g -O2 -I../include/ -fPIC -Wall -Wextra -Werror \
```
marmarek, line 30: Those cflags are needed for `shmoverride.o` too. A more flexible approach would be to keep CFLAGS variable for standard flags, and have extra_cflags for indeed local cflags (like the -I you wanted to add). Maybe call it user_cflags? Or... keep `CFLAGS+=` like it was before and then you can provide them with `CFLAGS=... make` call.

Similar comment applies to `gui-daemon/Makefile`.

marmarek (changes requested):

marmarek (approved):

PR #67, opened 2021-06-27
Use XCB instead of Xlib for shared memory

DemiMarie:
Xlib does not support file descriptor passing or dma-bufs.  The GUI daemon does not currently use either of these, but it will in the future.  Switching from Xlib to XCB is a necessary first step in that direction.

This is a subset of #65: it includes the parts that can work without needing any changes to the X server or kernel.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -643,6 +643,11 @@ static void mkghandles(Ghandles * g)
     }
     g->screen = DefaultScreen(g->display);
     g->root_win = RootWindow(g->display, g->screen);
+    g->gc = xcb_generate_id(g->cb_connection);
+    const xcb_void_cookie_t cookie = check_xcb_void(
+        xcb_create_gc_aux_checked(
+            g->cb_connection, g->gc, g->root_win, 0, NULL),
+        "xcb_create_gc_aux_checked");
     if (!XGetWindowAttributes(g->display, g->root_win, &attr)) {
         fprintf(stderr, "Cannot query window attributes!\n");
         exit(1);
@@ -653,6 +658,10 @@ static void mkghandles(Ghandles * g)
     g->clipboard_requested = 0;
     g->clipboard_xevent_time = 0;
     intern_global_atoms(g);
+    if (!g->context || xcb_request_check(g->cb_connection, cookie)) {
+        fprintf(stderr, "Failed to create global graphics context!\n");
+        exit(1);
+    }
     if (!XQueryExtension(g->display, "MIT-SHM",
                 &g->shm_major_opcode, &ev_base, &err_base))
         fprintf(stderr, "MIT-SHM X extension missing!\n");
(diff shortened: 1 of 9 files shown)


gui-daemon/xside.c
```diff
@@ -3004,65 +2998,66 @@ static void handle_window_dump(Ghandles *g, struct windowdata *vm_window,
...
+    if (error) {
         fprintf(stderr,
-            "XShmAttach failed for window 0x%lx(remote 0x%lx)\n",
+            "xcb_shm_attack_checked failed for window 0x%lx(remote 0x%lx)\n",
             vm_window->local_winid,
             vm_window->remote_winid);
-        XDestroyImage(vm_window->image);
-        vm_window->image = NULL;
-        shmctl(vm_window->shminfo.shmid, IPC_RMID, 0);
+        // shmctl(vm_window->shmid, IPC_RMID, 0);
```
marmarek, line 3048: This should not be commented out.

gui-daemon/xside.h
```diff
@@ -123,6 +128,8 @@ struct windowdata {
                                           request - translate it back when WM
                                           acknowledge maximize */
     uint32_t flags_set;    /* window flags acked to gui-agent */
+    uint32_t offset;       /**< Offset that X server should call mmap() at */
+    uint32_t grants;       /**< Number of grants mapped */
```
marmarek, line 132: Those two are unused, and likely won't be used with DMA-BUFs.

gui-daemon/xside.h
```diff
@@ -113,8 +118,8 @@ struct windowdata {
     struct windowdata *parent;    /* parent window */
     struct windowdata *transient_for;    /* transient_for hint for WM, see http://tronche.com/gui/x/icccm/sec-4.html#WM_TRANSIENT_FOR */
     int override_redirect;    /* see http://tronche.com/gui/x/xlib/window/attributes/override-redirect.html */
-    XShmSegmentInfo shminfo;    /* temporary shmid; see shmoverride/README */
-    XImage *image;        /* image with window content */
+    xcb_shm_seg_t shmseg; /* X Shared Memory segment */
+    int shmid;           /**< System V IPC identifier, or -1 for fd-passing */
```
marmarek, line 122: The -1 case is not applicable in this version. But it may be, so it's ok to leave this comment (and actual handling of -1 in few places).

marmarek (changes requested):

DemiMarie:
Everything I have tried works well.  Tray icons are not tested yet, and a Windows 7 qube keeps trying to snap back to full screen.  Not sure if that is normal.

marmarek:
> Windows 7 qube keeps trying to snap back to full screen. Not sure if that is normal.

It may be the case, yes.

marmarek (approved):

DemiMarie:
Tray icons work.

PR #69, opened 2021-06-28
Use dma-bufs for shared memory with X

DemiMarie:
See commit messages for details.


DemiMarie:
I tried to test this and ran into a problem with our kernel configuration: XEN_GNTDEV_DMABUF is not set.

DemiMarie:
CI is failing because the CentOS 8 kernel headers do not define the needed ioctls.  Should I redefine them myself?

marmarek:
> CI is failing because the CentOS 8 kernel headers do not define the needed ioctls. Should I redefine them myself?

@fepitre is CentOS going to work as GUI domain? If no, don't bother @DemiMarie  and simply exclude CentOS in the `Makefile.builder`, like at https://github.com/QubesOS/qubes-python-u2flib-host/commit/fad317e1fee7d98d54301905bab8c7e407742a3e

fepitre:
> > CI is failing because the CentOS 8 kernel headers do not define the needed ioctls. Should I redefine them myself?
> 
> @fepitre is CentOS going to work as GUI domain? If no, don't bother @DemiMarie and simply exclude CentOS in the `Makefile.builder`, like at [QubesOS/qubes-python-u2flib-host@fad317e](https://github.com/QubesOS/qubes-python-u2flib-host/commit/fad317e1fee7d98d54301905bab8c7e407742a3e)

It should but don't bother with that right now. If any easier solution comes later, we can enable it.

DemiMarie:
I was trying to figure out why I kept getting `-EINVAL` from the kernel.  There was a check that caused `IOCTL_GNTDEV_DMABUF_EXP_FROM_REF` to always fail with `-EINVAL` in a PV domain.  I removed the check and rebuilt the kernel, and then applied the following patch:

```patch
diff --git a/shmoverride/shmoverride.c b/shmoverride/shmoverride.c
index a96fe7f..fb85367 100644
--- a/shmoverride/shmoverride.c
+++ b/shmoverride/shmoverride.c
@@ -111,12 +111,26 @@ static uint8_t *shmat_grant_refs(struct shm_args_hdr *shm_args,
 
     info->u.grant.count = shm_args_grant->count;
 
+    uint32_t fd;
+    if (!xengnttab_dmabuf_exp_from_refs(xgt,
+            shm_args->domid,
+            0,
+            shm_args_grant->count,
+            &shm_args_grant->refs[0],
+            &fd)) {
+        fputs("xengnttab_dmabuf_exp_from_refs worked\n", stderr);
+        close(fd);
+    } else {
+        fputs("xengnttab_dmabuf_exp_from_refs failed\n", stderr);
+    }
+    fflush(stderr);
     map = xengnttab_map_domain_grant_refs(xgt,
             shm_args_grant->count,
             shm_args->domid,
             &shm_args_grant->refs[0],
             PROT_READ);
 
+    fprintf(stderr, "Map returned address %p\n", map);
     return map;
 }
 
```

Again, `xengnttab_dmabuf_exp_from_refs` (which uses `IOCTL_GNTDEV_DMABUF_EXP_FROM_REF`) failed with `-EINVAL`, indicating that the parameters are incorrect.

But we know the parameters are correct!  If they were wrong, the following call to `xengnttab_map_domain_grant_refs` would fail in the unpatched version, meaning that shmoverride (and the GUI daemon) could not possibly work.  And it does work.

Conclusion: Kernel bug.

DemiMarie:
Turns out that grant table dma-bufs don’t even support `mmap()`!

DemiMarie:
Closing for now, as this would require kernel patches to work at all, and even more patches to work in dom0.  This can be changed if and when support for PVH dom0 arrives.

PR #71, opened 2021-07-01
force_on_screen(): fix calculations

DemiMarie:
Instead of trying to check for offscreen windows while sanitizing window
positions, return early in the case of offscreen windows, and then
sanitize x, y, width, height, x + width, and y + height separately.
This takes more code, but the new code is much, *much* simpler and
easier to understand.

Fixes QubesOS/qubes-issues#6518.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -429,8 +429,14 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
     return child_win;
 }
 
-static const int desktop_coordinates_size = 4;
+static const unsigned long desktop_coordinates_size = 4;
 static const long max_display_width = 1UL << 20;
+
+/*
+ * Padding enforced between override-redirect windows and the edge of
+ * the work area.
+ */
+static const int override_redirect_padding = 0;
 
 /* update g when the current desktop changes */
 static void update_work_area(Ghandles *g) {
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -1409,43 +1409,71 @@ static int force_on_screen(Ghandles * g, struct windowdata *vm_window,
...
+        (x < 0 && x + w < 0) ||
+        (y < 0 && w + h < 0)) {
+        /* Window is entirely off-screen (the border does not matter here) */
+        return 0;
+    }
+
+    if (g->work_width < 2 * border_width ||
+        g->work_height < 2 * border_width) {
+        /* Work area too small for a border??? */
+        vm_window->x = vm_window->y = vm_window->width = vm_window->height = 0;
```
marmarek, line 1428: Size must be non-zero, you'll get BadValue error from X server otherwise.

gui-daemon/xside.c
```diff
@@ -1409,43 +1409,71 @@ static int force_on_screen(Ghandles * g, struct windowdata *vm_window,
...
         border_y = border_width + g->work_y,
         max_width = g->work_width - border_width,
         max_height = g->work_height - border_width;
-    if (vm_window->x < border_x
-        && vm_window->x + border_width + (int)vm_window->width > 0) {
+    /* Sanitize width and height */
+    if (w > max_width) {
+        vm_window->width = max_width;
+        do_move = 1;
+        reason = 5;
```
marmarek, line 1440: Make it a bitmask? Now it's even more likely that several of those will apply at the same time.

gui-daemon/xside.c
DemiMarie: Is it okay to just exit in this absurd situation?  Not to mention that this code is currently unreachable.

gui-daemon/xside.c
DemiMarie: Done

DemiMarie:
PipelineRetry

marmarek:
PipelineRetry

marmarek (approved):

PR #73, opened 2021-07-04
Check if a window is referenced by another one before destroying it

marmarek:
There can be two types of relations:
- `parent`: in this case agent must destroy children windows first - if
 that isn't the case, terminate abruptly
- `transient_for`: in that case clear the reference (both in local
 structure and in the X server)

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2704,8 +2704,12 @@ static void handle_map(Ghandles * g, struct windowdata *vm_window)
         vm_window->transient_for = transdata;
         XSetTransientForHint(g->display, vm_window->local_winid,
                      transdata->local_winid);
-    } else
+    } else {
+        if (vm_window->transient_for)
+            XDeleteProperty(g->display, vm_window->local_winid,
+                    XA_WM_TRANSIENT_FOR);
         vm_window->transient_for = NULL;
+    }
 
     set_override_redirect(g, vm_window, !!(untrusted_txt.override_redirect));
     attr.override_redirect = vm_window->override_redirect;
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -2201,6 +2201,36 @@ static void handle_create(Ghandles * g, XID window)
...
+{
+    struct genlist *iter;
+
+    list_for_each(iter, g->wid2windowdata) {
+        struct windowdata *iter_window = iter->data;
+        if (iter_window->parent == vm_window) {
+            fprintf(stderr, "Window 0x%x is still a parent for 0x%x, "
+                    "but VM tried to destroy it\n",
+                   (int)vm_window->local_winid, (int)iter_window->local_winid);
+            fprintf(stderr, "Aborting\n");
```
DemiMarie, line 2220:
```suggestion
                    "but VM tried to destroy it\nAborting\n",
                   (int)vm_window->local_winid, (int)iter_window->local_winid);
```

gui-daemon/xside.c
```diff
@@ -2201,6 +2201,36 @@ static void handle_create(Ghandles * g, XID window)
...
+            exit(1);
+        }
+        if (iter_window->transient_for == vm_window) {
+            fprintf(stderr, "Window 0x%x is still set as transient_for "
+                    "for a 0x%x window, but VM tried to destroy it\n",
```
DemiMarie, line 2225:
```suggestion
                    "for window 0x%x, but VM tried to destroy it\n"
                    "Unsetting transient_for property.\n",
```

DemiMarie (commented):
Tiny nits

PR #74, opened 2021-07-05
Do not draw borders around subwindows

DemiMarie:
A subwindow is guaranteed to be contained in its parent window, so there
is no need to draw a border around it.

Must not be merged before #73.


DemiMarie:
@marmarek this can be merged now.

gui-daemon/xside.c
```diff
@@ -1779,6 +1779,17 @@ static void do_shm_update(Ghandles * g, struct windowdata *vm_window,
     if (!vm_window->override_redirect) {
         // Window Manager will take care of the frame...
         border_width = 0;
+    } else if (vm_window->parent) {
+        if (x >= vm_window->parent->x &&
+            y >= vm_window->parent->y &&
+            (int64_t)w + x <= (int64_t)vm_window->parent->width +
+                              vm_window->parent->x &&
+            (int64_t)h + y <= (int64_t)vm_window->parent->height +
+                              vm_window->parent->y) {
```
marmarek, line 1788: The coordinates of a window are relative to its parent (which is the root window if `parent == NULL`), so this check is wrong.
I believe our gui-agent-linux never sends windows with `parent != NULL`, so it may be not easy to test. Not sure about the gui-agent-window. Furthermore, I believe override-redirect makes no sense (as in: is no-op) for non-top-level windows, so this case is unlikely happen in practice - but I'm not 100% sure about this one.

marmarek (changes requested):
Since this doesn't apply to the current X11 clients, and also it doesn't fix missing borders case (but rather: extra borders), I'd wait with merging it until we see an actual use case for this (likely with Wayland gui-agent)

DemiMarie:
> Since this doesn't apply to the current X11 clients, and also it doesn't fix missing borders case (but rather: extra borders), I'd wait with merging it until we see an actual use case for this (likely with Wayland gui-agent)

:confused: I thought this would fix the FreeRDP rendering anomalies.

marmarek:
> confused I thought this would fix the FreeRDP rendering anomalies.

Haven you seen a gui-daemon (debug) log that confirms it indeed uses windows with non-root parent? I would be surprised, but it isn't totally impossible (and indeed would make this PR worth merging earlier).

DemiMarie:
> > confused I thought this would fix the FreeRDP rendering anomalies.
> 
> Haven you seen a gui-daemon (debug) log that confirms it indeed uses windows with non-root parent? I would be surprised, but it isn't totally impossible (and indeed would make this PR worth merging earlier).

I would be surprised if that *was not* the case, but I also do not have any way to reproduce the issue myself.

DemiMarie:
I wonder if this fixes the warning I have been getting from LibreOffice.

DemiMarie:
This would create a security vulnerability if merged, as it allows a client to overwrite its own window.

DemiMarie:
There are several approaches for fixing the vulnerability:

1. Draw the border on the outside of the window instead of the inside.  This is somewhat complex, as it affects the window size and thus positioning.
2. Only skip drawing the border if the outermost window is _not_ an override redirect window.
3. Only skip drawing the border if the window is more than `border_width` inside the outermost window.

PR #78, opened 2021-07-10
Use FD passing for shared memory

DemiMarie:
This is another attempt at using FD passing for shared memory, at the cost of still requiring shmoverride.
    
~~This doesn’t actually work, and it may not be a good use of time even if it did.  The reason I am filing it as a PR is to at least give it a chance to be reviewed, so that the work done on it has a chance of not being wasted.~~  The code is now fully functional and has been tested over VNC.  The main advantage is that the same approach works on Wayland, which does not support System V IPC.  Newer versions of glibc (which do not use the legacy `__fxstat` and `__fxstat64` functions) are known to build but have not been otherwise tested.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -10,9 +10,10 @@ Build-Depends:
  libvchan-xen-dev,
  python3-dev,
  libpulse-dev,
- libxext-dev,
  libxrandr-dev,
  libxcb1-dev,
+ libxcb-util0-dev,
+ libxcb-shm0-dev,
  libx11-xcb-dev,
  libconfig-dev,
  libpng-dev,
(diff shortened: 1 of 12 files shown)


DemiMarie:
This actually works now!  I am still not sure if the extra complexity (and overhead) is worthwhile, but this shows that the approach is viable.  Tray icons have not been tested.

DemiMarie:
Last few commits add some consistency checks and fix a premature free() in shmoverride, which led to every element of shmoverride’s global list being located in freed memory.

DemiMarie:
PipelineRetry

DemiMarie:
This has passed every test I can think of, which proves that shmoverride will not be needed at all for Wayland.

DemiMarie:
PipelineRefresh

DemiMarie:
Windows qubes work, too!

marmarek:
I'm going to leave it open until after branching off R4.1.

DemiMarie:
PipelineRetry

DemiMarie:
@marmarek please re-review

marmarek:
GUI daemon crashes with `qubes-guid: XGetWindowAttributes: Inappropriate ioctl for device`
You can extract more logs from var_log.tar.gz in https://openqa.qubes-os.org/tests/43364#downloads

DemiMarie:
> GUI daemon crashes with `qubes-guid: XGetWindowAttributes: Inappropriate ioctl for device` You can extract more logs from var_log.tar.gz in https://openqa.qubes-os.org/tests/43364#downloads

That was due to an incorrect check for the return value of `XGetWindowAttributes` in an old version of #105.  I fixed the problem in that PR.  Please re-enable openQA on this PR.

DemiMarie:
Closing in favor of #116.

PR #80, opened 2021-07-18
Unset fullscreen when a window leaves focus

DemiMarie:
This is a security measure, to ensure that a window can be forced out of the fullscreen state.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -580,8 +580,8 @@ static void intern_global_atoms(Ghandles *const g) {
         *atoms_to_intern[i].dest = labels[i];
 }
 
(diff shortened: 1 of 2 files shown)


DemiMarie:
I marked this as a draft as it does not properly handle multiple monitors.  Specifically:

1. A fullscreen window should *stay* fullscreen even if the user uses the system normally on another monitor.
2. The `_NET_ACTIVE_WINDOW` property appears to be multi-valued, presumably because of multiple monitor support.

DemiMarie:
PipelineRetry

DemiMarie:
As per discussion with @marmarek and @ninavizz, I am closing this PR in favor of a better solution that does not break multi-monitor support.

PR #81, opened 2021-08-05
Add comments in do_shm_update

DemiMarie:
This code is incredibly tricky and difficult to understand.  It turns
out to be correct, but checking that took hours, so add some comments to
help anyone else who looks at it.


DemiMarie:
There were at least two times when reading this code that I thought it had a security vulnerability.  That is how tricky this is.

DemiMarie:
PipelineRetry

DemiMarie:
PipelineRetry

marmarek:
PipelineRetry

PR #82, opened 2021-08-17
Draft: Consider work area when positioning windows

DemiMarie:
Marking as draft because I am not sure if the call to abort() is
justified.

Fixes QubesOS/qubes-issues#6518.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -1590,14 +1590,14 @@ static int force_on_screen(Ghandles * g, struct windowdata *vm_window,
         reason |= TOP_BORDER_OFF_SCREEN;
     }
     /* Sanitize right */
-    if (vm_window->x > max_width - (int)vm_window->width) {
-        vm_window->x = max_width - (int)vm_window->width;
+    if (vm_window->x > border_x + (max_width - (int)vm_window->width)) {
+        vm_window->x = border_x + (max_width - (int)vm_window->width);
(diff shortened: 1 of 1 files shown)


marmarek (changes requested):
No, `abort()` definitely is not warranted in this case.

Plus, this change doesn't doesn't fix the bug. It introduces another one - you are now resizing the window (why?) while the original issue was about the window position, not size.

marmarek:
I think the issue is here: https://github.com/QubesOS/qubes-gui-daemon/blob/master/gui-daemon/xside.c#L1599-L1600
```
    if (vm_window->y > max_height - (int)vm_window->height) {
        vm_window->y = max_height - (int)vm_window->height;
```
Shouldn't this be `border_y + max_height - (int)vm_window->height` (and similar for right border)?

DemiMarie:
> I think the issue is here: https://github.com/QubesOS/qubes-gui-daemon/blob/master/gui-daemon/xside.c#L1599-L1600
> 
> ```
>     if (vm_window->y > max_height - (int)vm_window->height) {
>         vm_window->y = max_height - (int)vm_window->height;
> ```
> 
> Shouldn't this be `border_y + max_height - (int)vm_window->height` (and similar for right border)?

Indeed, oops.  That said, I do believe that `moveresize_vm_windows` also has a bug.

marmarek (approved):

PR #84, opened 2021-09-26
Fix shmoverride permissions

DemiMarie:
Trivial change but needs testing.


marmarek (changes requested):
Please add some more context to the commit message - who+where creates this SHM segment and with what permissions.

DemiMarie:
> Please add some more context to the commit message - who+where creates this SHM segment and with what permissions.

Nevermind, this patch is wrong.

PR #85, opened 2021-09-28
Fix non-refreshed tray icons

marmarek:
Force refreshing tray icon just after docking it

Fixes QubesOS/qubes-issues#6856

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2368,7 +2368,7 @@ static void handle_shmimage(Ghandles * g, struct windowdata *vm_window)
     struct msg_shmimage untrusted_mx;
 
     read_struct(g->vchan, untrusted_mx);
-    if (!vm_window->is_mapped)
+    if (!vm_window->is_mapped && !vm_window->is_docked)
         return;
     if (g->log_level >= 2) {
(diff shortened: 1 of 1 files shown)


marmarek:
With what application? nm-applet?

jevank:
tested with nm-applet

jevank:
Updated: better behavior - it refreshed by mouse over

marmarek:
It looks like this is actually a bug on the VM side, uncovered by this change. Can you test QubesOS/qubes-gui-agent-linux#144 ?

jevank:
Nope, still refreshed with mouse over only. I tested with nm-applet and see blank area first and then see no animation while wifi connection after manipulations, usually it has.

marmarek:
Ok, I think I've found what is wrong (see commit messages). The issue was much simpler and does explain why it has changed with QSB-072 patches. Can you test?

jevank:
Yes, that is the fix. nm-applet animation is back.

noskb:
I've tested it on R4.0 and can confirm the fix. 
`/var/log/qubes/guid.sys-whonix.log`
```
Icon size: 128x128
docking an override-redirect window 0x4400006 - clearing override-redirect
```

PR #86, opened 2021-10-05
Disable subwindows by default

DemiMarie:
The default GUI agent never creates them, and they are extra attack surface.  For instance, they are involved in one of the exploits for QSB#072.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2459,6 +2459,12 @@ static void handle_create(Ghandles * g, XID window)
     /* prevent using a parent that is either override-redirect, or docked */
     if (l) {
         struct windowdata *vm_window_parent = l->data;
+        if (!g->permit_subwindows) {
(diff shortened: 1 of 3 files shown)


gui-daemon/xside.c
```diff
@@ -2406,6 +2406,10 @@ static void handle_create(Ghandles * g, XID window)
      */
     read_struct(g->vchan, untrusted_crt);
     /* sanitize start */
+    if (untrusted_crt.parent && !g->permit_subwindows) {
```
marmarek, line 2409: This is wrong check, because the gui-agent (unfortunately) sends the root window ID here for top-level windows.  You need to move it after the lookup of that window (see where I added the check for override-redirect subwindows).

DemiMarie:
@marmarek would you mind re-reviewing this?

marmarek (changes requested):
Please add the option also to `guid.conf`, which serves as a documentation of supported options.

marmarek (approved):

PR #87, opened 2021-10-09
Disable child windows entirely

DemiMarie:
X11 child windows are not capable enough to implement Wayland subsurfaces, so Wayland subsurfaces will need to be software composited anyway.  This leaves child windows with no uses, so ban them.


DemiMarie:
Closing in favor of #86.

PR #93, opened 2022-02-23
Send xinput events

locriacyber:
https://github.com/QubesOS/qubes-issues/issues/7231


locriacyber:
Question: how do I inject raw events and mapped events independently?

PR #94, opened 2022-02-24
Don't send repeat key presses to clients

locriacyber:
Attempt to fix https://github.com/QubesOS/qubes-issues/issues/7231 partially.

Not tested due to GLIBC version mismatch.


locriacyber:
New problem: fedora32 templates don't generate key repeat events, while archlinux do.

locriacyber:
> Check if it isn't disabled via xset.
I enabled it via xset and it now works.
I think just dropping repeated key events is better. The client OS should generate its own key repeat events.
I think archlinux templates will generate its own set of repeat events (so double the repeat speed). I really don't know what

locriacyber:
Also, building this in fedora 32 now requires package `libXi-devel` (for CI to pass). How do I add that?

marmarek:
I mean, send events including the (repeat) flag and let the gui agent decide. It can either:
1. ignore repeat events and let local X server generate them
2. disable repeat in local X server and use those sent by gui-daemon

Both have some pros and cons, with 2nd option respecting your globally configured repeat speed - that's why I think it is a better solution. But if going with the former, it should be at least opt-in (can be via config option), or combined with some mechanism ensuring VM actually have its own key repeat enabled (probably using qvm-feature mechanism). Otherwise it will break some existing configurations.

> Also, building this in fedora 32 now requires package `libXi-devel` (for CI to pass). How do I add that?

Add to `rpm_spec/gui-daemon.spec.in`, and also to `debian/control`.

locriacyber:
New problem: when switching windows from window A to window B using "Super + Tab", the tab key press was sent to window A.

locriacyber:
This solution doesn't work well. I'll try using XInput for all input events instead.

PR #97, opened 2022-03-04
Send XInput events to VM

locriacyber:
Attempt to Close https://github.com/QubesOS/qubes-issues/issues/7316, Close https://github.com/QubesOS/qubes-issues/issues/7231

TODO:
- [ ] intercept clipboard shortcuts
- [ ] create fake device id; current implementation may result in device id collision in client side
- [ ] patch linux agent (note to self: https://gist.github.com/locriacyber/937c013ce8175016694b1b2907ac00fe)


locriacyber:
This is incorrect. A better way have one process and only send raw input events to one window at once.

marmarek:
I don't like what is happening here.
As explained in an email, receiving events even without having focus and then (trying to) track focus independently for filtering is too risky. One event missed/mishandled/whatever and you have cross-vm information leak.
Additionally, sending `XIDeviceEvent` as is, as part of the GUI protocol is no-go. Among other things, GUi protocol definition must _not_ depend on X headers.

Lets start from the goal you want to achieve (mouse events usable for games?), instead of proxying whatever X11 normally does. X11 is really not a good source for secure GUI protocol design...

By reading https://github.com/QubesOS/qubes-issues/issues/7316, it seems the issue is purely on gui-agent side, about how events are reported/injected to the VM's X server.

locriacyber:
> Additionally, sending XIDeviceEvent as is, as part of the GUI protocol is no-go. Among other things, GUi protocol definition must not depend on X headers.
This is a prototype that tries to be work correctly. I didn't bothered to create another struct because I need to send ~80% what `XIDeviceEvent` has.

> By reading https://github.com/QubesOS/qubes-issues/issues/7316, it seems the issue is purely on gui-agent side, about how events are reported/injected to the VM's X server.
It's not purely on gui-agent's side. The VM get fed key repeats by dom0.

DemiMarie:
@locriacyber @marmarek Can this be closed in favor of #98?

locriacyber:
Yes. Closing. See #98 for working (mostly) version.

PR #98, opened 2022-03-05
Drop repeat key events

locriacyber:
This is actually usable now.
Close https://github.com/QubesOS/qubes-issues/issues/7231


gui-daemon/xside.c
```diff
@@ -1931,15 +1946,6 @@ static void process_xevent_focus(Ghandles * g, const XFocusChangeEvent * ev)
     struct msg_focus k;
     CHECK_NONMANAGED_WINDOW(g, ev->window);
 
-    /* Ignore everything other than normal, non-temporary focus change. In
-     * practice it ignores NotifyGrab and NotifyUngrab. VM does not have any
-     * way to grab focus in dom0, so it shouldn't care about those events. Grab
-     * is used by window managers during task switching (either classic task
-     * switcher, or KDE "present windows" feature).
-     */
-    if (ev->mode != NotifyNormal && ev->mode != NotifyWhileGrabbed)
```
locriacyber, line 1940: Turns out this is the reason for extraneous keystrokes when cycling windows.
The window manager use `NotifyGrab` to signal temponany focus out.

locriacyber:
Actually, there is no telling that this "keypress" would come from a keyboard or not.

gui-daemon/xside.c
```diff
@@ -2318,66 +2324,127 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
-                         event_buffer.xclient.window);
+    if (XGetEventData(g->display, cookie) &&
+            cookie->type == GenericEvent &&
+            cookie->extension == g->xi_opcode) {
+        XIEvent* xi_event = cookie->data; // from test_xi2.c in xinput cli utility
+
+        XIDeviceEvent * xi_device = (XIDeviceEvent *)xi_event;
+        XILeaveEvent * xi_leave = (XILeaveEvent *)xi_event;
+        switch (xi_event->evtype) {
+        // ideally raw input events are better, but I'm relying on X server's built-in event filtering and routing feature here
```
DemiMarie, line 2389: Is this to avoid information leaks if the GUI daemon thinks it has focus when it does not?

gui-daemon/xside.c
```diff
@@ -2318,66 +2324,127 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+        case XI_KeyPress:
+        case XI_KeyRelease:
+            if (xi_device && xi_device->flags & XIKeyRepeat) break; // don't send key repeat events
```
DemiMarie, line 2392:
```suggestion
            if (xi_device && (xi_device->flags & XIKeyRepeat)) break; // don't send key repeat events
```

gui-daemon/xside.c
```diff
@@ -2318,66 +2324,127 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+            process_xievent_keypress(g, xi_device);
+            break;
+        case XI_FocusIn:
+        case XI_FocusOut:
+            process_xievent_focus(g, xi_leave);
```
DemiMarie, line 2397:
```suggestion
            process_xievent_focus(g, (XILeaveEvent *)xi_event);
```

gui-daemon/xside.c
```diff
@@ -2318,66 +2324,127 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
-                    (int) event_buffer.xclient.window);
-            process_xevent_close(g,
-                         event_buffer.xclient.window);
+    if (XGetEventData(g->display, cookie) &&
+            cookie->type == GenericEvent &&
+            cookie->extension == g->xi_opcode) {
+        XIEvent* xi_event = cookie->data; // from test_xi2.c in xinput cli utility
+
+        XIDeviceEvent * xi_device = (XIDeviceEvent *)xi_event;
+        XILeaveEvent * xi_leave = (XILeaveEvent *)xi_event;
```
DemiMarie, line 2387:
```suggestion
```

gui-daemon/xside.c
DemiMarie: I kept having problems with this when writing the Wayland compositor.  I had no idea what the cause was, though.  I wound up having to try different approaches until I found one that worked.  Hopefully this PR will fix the actual problem.

gui-daemon/xside.c
```diff
@@ -374,10 +376,23 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
...
+    // select xinput events
+    XIEventMask xi_mask;
+    xi_mask.deviceid = XIAllDevices;
+    xi_mask.mask_len = XIMaskLen(XI_LASTEVENT);
+    xi_mask.mask = calloc(xi_mask.mask_len, sizeof(char));
+    XISetMask(xi_mask.mask, XI_KeyPress);
+    XISetMask(xi_mask.mask, XI_KeyRelease);
+    XISetMask(xi_mask.mask, XI_FocusIn);
+    XISetMask(xi_mask.mask, XI_FocusOut);
+    XISelectEvents(g->display, child_win, &xi_mask, 1);
```
DemiMarie, line 393: Do any of these need their return code checked?

locriacyber: `XISelectEvents` may generate error (call callback), but I'm not sure what the return value means.

https://www.x.org/releases/current/doc/libXi/inputlib.html#XSelectExtensionEvent

The documentation is outdated.

gui-daemon/xside.c
locriacyber: I still don't know how to send correct focus events to windows. I tried:
- `XSendEvent`
- `XGrabKeyboard`
I don't know about Wayland though, I only tested with X server in VM.

gui-daemon/xside.c
locriacyber: Raw input events are sent to the root window, not individual windows. If I want to use raw events, I need to track window focus myself. Also, raw key events don't have modifiers key state (`Ctrl`, `Alt`), so I'll have to track that too.

gui-daemon/xside.c
DemiMarie: Looks like the only way to find out is to check the C source code :(

gui-daemon/xside.c
DemiMarie: >   I don't know about Wayland though, I only tested with X server in VM.

Don’t worry about that.  Something that works properly under X is hopefully going to be easier to handle with Wayland as well.

DemiMarie:
Code-signing check is failing because GitHub’s key isn’t considered verified.

gui-daemon/xside.c
```diff
@@ -2318,66 +2325,126 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
-                fprintf(stderr, "close for 0x%x\n",
-                    (int) event_buffer.xclient.window);
-            process_xevent_close(g,
-                         event_buffer.xclient.window);
+    if (XGetEventData(g->display, cookie) &&
+            cookie->type == GenericEvent &&
+            cookie->extension == g->xi_opcode) {
+        XIEvent* xi_event = cookie->data; // from test_xi2.c in xinput cli utility
+
+        XIDeviceEvent * xi_device = (XIDeviceEvent *)xi_event;
```
DemiMarie, line 2387:
```suggestion
```

gui-daemon/xside.c
```diff
@@ -2318,66 +2325,126 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+        switch (xi_event->evtype) {
+        // ideally raw input events are better, but I'm relying on X server's built-in event filtering and routing feature here
+        case XI_KeyPress:
+        case XI_KeyRelease:
+            if (xi_device && (xi_device->flags & XIKeyRepeat)) break; // don't send key repeat events
```
DemiMarie, line 2392:
```suggestion
```

Key repeats should be ignored in `process_xievent_keypress`, as otherwise the screen will lock prematurely if one key repeats for a very long time.

gui-daemon/xside.c
```diff
@@ -374,10 +376,25 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
...
                 ButtonPressMask | ButtonReleaseMask |
                 PointerMotionMask | EnterWindowMask | LeaveWindowMask |
-                FocusChangeMask | StructureNotifyMask | PropertyChangeMask);
+                StructureNotifyMask | PropertyChangeMask);
+    
+    // select xinput events
+    XIEventMask xi_mask;
+    xi_mask.deviceid = XIAllMasterDevices; // https://stackoverflow.com/questions/44095001/getting-double-rawkeypress-events-using-xinput2
+    xi_mask.mask_len = XIMaskLen(XI_LASTEVENT);
+    xi_mask.mask = calloc(xi_mask.mask_len, sizeof(char));
```
DemiMarie, line 388:
```suggestion
    if (!(xi_mask.mask = calloc(xi_mask.mask_len, sizeof(char)))) {
        fputs("Out of memory!\n", stderr);
        exit(1);
    }
```

missing NULL check

gui-daemon/xside.c
```diff
@@ -647,6 +664,9 @@ static void mkghandles(Ghandles * g)
     if (!XQueryExtension(g->display, "MIT-SHM",
                 &g->shm_major_opcode, &ev_base, &err_base))
         fprintf(stderr, "MIT-SHM X extension missing!\n");
+    if (!XQueryExtension(g->display, "XInputExtension", &g->xi_opcode, &ev_base, &err_base)) {
+        fprintf(stderr, "X Input extension not available. Key press events not available. Upgrade your X11 server now.\n");
```
DemiMarie, line 668:
```suggestion
        fprintf(stderr, "X Input extension not available. Key press events not available. Upgrade your X11 server now.\n");
        exit(1);
```

I don’t think it is worth continuing in this case, especially since it indicates a packaging bug.  The daemon should depend on a sufficiently new X server to prevent this.

gui-daemon/xside.c
```diff
@@ -1329,22 +1349,22 @@ static void update_wm_user_time(Ghandles *const g, const Window window,
...
  */
-static void process_xevent_keypress(Ghandles * g, const XKeyEvent * ev)
+static void process_xievent_keypress(Ghandles * g, const XIDeviceEvent * ev)
 {
     struct msg_hdr hdr;
     struct msg_keypress k;
-    CHECK_NONMANAGED_WINDOW(g, ev->window);
-    update_wm_user_time(g, ev->window, ev->time);
+    CHECK_NONMANAGED_WINDOW(g, ev->event);
+    update_wm_user_time(g, ev->event, ev->time);
```
DemiMarie, line 1360:
```suggestion
    // yes, ev->event is the window number
    update_wm_user_time(g, ev->event, ev->time);
    if (ev->flags & XIKeyRepeat)
        return; // don't send key repeat events    
```

It’s probably best to call `update_wm_user_time` even if no event will be sent, since a key repeat should still prevent screen lock.  Otherwise the screen would lock too soon if a user held down a key for a long time.

gui-daemon/xside.c
```diff
@@ -2318,66 +2325,126 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+//     fake_event.x = xi_event->event_x;
+//     fake_event.y = xi_event->event_y;
+//     fake_event.x_root = xi_event->root_x;
+//     fake_event.y_root = xi_event->root_y;
+//     fake_event.state = xi_event->mods.effective;
+//     fake_event.keycode = xi_event->detail;
+//     fake_event.same_screen = true; // don't know how to fill this
+//     return fake_event;
+// }
+
```
DemiMarie, line 2375:
```suggestion
```

Commented-out code is fine in development, but it should not make it into production code.

DemiMarie (changes requested):
Thanks for the patch!  This should fix some _really_ ugly bugs when it comes to window switching.  Avoiding them required a disgusting hack in the Wayland GUI agent that I still do not really understand.

Some suggestions below:

DemiMarie:
Also, don’t forget to add “FIxes: QubesOS/qubes-issues#7396” to whichever commit fixed that bug.

locriacyber:
I've used it for a month, and it's pretty stable. Sometimes, shortcuts like Super+S to open windows in dom0 will get S key stuck (likely focus issue).

DemiMarie (approved):

marmarek:
It doesn't build, see CI logs:
```
DEBUG: make[1]: Entering directory '/builddir/build/BUILD/qubes-gui-daemon-4.1.21/gui-daemon'
DEBUG: gcc -MD -MP -MF xside.o.dep -c -o xside.o -I../include/ -g -O2 -Wall -Wextra -Werror -pie -fPIC -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/vchan-xen -I/usr/include/libpng16 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/libmount -I/usr/include/blkid -pthread  -fvisibility=hidden -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection xside.c
DEBUG: xside.c: In function 'process_xevent':
DEBUG: xside.c:2359:41: error: 'xi_device' undeclared (first use in this function)
DEBUG:  2359 |             process_xievent_keypress(g, xi_device);
DEBUG:       |                                         ^~~~~~~~~
DEBUG: xside.c:2359:41: note: each undeclared identifier is reported only once for each function it appears in
DEBUG: make[1]: Leaving directory '/builddir/build/BUILD/qubes-gui-daemon-4.1.21/gui-daemon'
```

marmarek:
Please also squash git history into functional commits (no WIP commits).

gui-daemon/xside.c
```diff
@@ -2322,62 +2336,72 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    if (XGetEventData(g->display, cookie) &&
+            cookie->type == GenericEvent &&
+            cookie->extension == g->xi_opcode) {
+        XIEvent* xi_event = cookie->data; // from test_xi2.c in xinput cli utility
+
+        switch (xi_event->evtype) {
+        // ideally raw input events are better, but I'm relying on X server's built-in event filtering and routing feature here
+        case XI_KeyPress:
+        case XI_KeyRelease:
+            process_xievent_keypress(g, xi_device);
```
DemiMarie, line 2350:
```suggestion
            process_xievent_keypress(g, (XIDeviceEvent *)xi_event);
```

DemiMarie (changes requested):
This should fix the build.

DemiMarie:
Is it possible to use libxcb instead of libXi?

locriacyber:
> Is it possible to use libxcb instead of libXi?
Yes, but need to change source code to use xcb functions.

locriacyber:
> Please also squash git history into functional commits (no WIP commits).

How do I do that? I feel like the commit history is descriptive enough.

marmarek:
> Is it possible to use libxcb instead of libXi?

Using libxcb and libX11-like together, both for handling events is a bad idea. You will run into events reordering (and probably _a lot_ of other issues). Not worth it in this PR.

marmarek:
> How do I do that? I feel like the commit history is descriptive enough.

I don't like introducing not working commit that is known to be not working at merge time already (that is later fixed or even reverted in the same PR).  For example https://github.com/QubesOS/qubes-gui-daemon/pull/98/commits/60ee007e74284a422cdbbe7ff71669d1cfd4f173 is later basically reverted. You included also some non-trivial changes into a _merge commit_ 6bddd0e04952a64d3bb4b482d07a7e1633da6ddc...
Can you rebase it and keep only commits that are relevant to the final version?

locriacyber:
I'm not good at git to know how to rebase something like (A -> B -> revert A). I also think that git commits should reflect how software is actually developed, not how it **should** be ([source](https://www.fossil-scm.org/home/doc/trunk/www/fossil-v-git.wiki#history)).

The diff of https://github.com/QubesOS/qubes-gui-daemon/commit/6bddd0e04952a64d3bb4b482d07a7e1633da6ddc is simple if you use a language-aware diff tool like https://github.com/Wilfred/difftastic.

Is it possible to squash the whole thing into a single commit?

marmarek:
> Is it possible to squash the whole thing into a single commit?

Take a look here for example: https://gitbetter.substack.com/p/how-to-squash-git-commits

Anyway, I tried to use it, and the key repeat in VM doesn't work. It isn't surprising, because it's disabled. After enabling it, it works. Just enabling it is not a problem, but I'm worried about extra parameters (delay, rate). Those user normally set in dom0 (or wherever GUI runs) and should apply to all the VMs. Synchronizing them may be messy but probably doable... Anyway, handling it in some way is necessary for merging this change.

marmarek:
> system_tests_gui_interactive
> 
>     * clipboard_and_web: [unnamed test](https://openqa.qubes-os.org/tests/40791#step/clipboard_and_web/16) (unknown)
> 
>     * clipboard_and_web: [Failed](https://openqa.qubes-os.org/tests/40791#step/clipboard_and_web/17) (test died)
>       `# Test died: no candidate needle with tag(s) 'qubes-website' matche...`

This failure actually may be related to this PR. The test is supposed to type in `https://www.qubes-os.org/`, but it ends up with `https://https://ww.qubes-os.org/`. I'm not sure why exactly one "w" is missing, but this failure can be reproduced rather reliably (I tried restarting the test several times - it happened in 4 out of 5 runs).

gui-daemon/xside.c
DemiMarie: `XISelectEvents()` can only fail if there is a memory allocation error, the XInput 2 extension is not present or is too old, or the arguments are invalid.  I don’t think any of these are worth trying to recover from, so just exit if the return value is not `Success`.

DemiMarie (changes requested):

locriacyber:
I've been using this for awhile now. Fixed the remaining bug.

There is still 2 last bugs:

1.
Sometimes, spawned window from XFCE4 Application launcher (dom0) will close itself (seem to be closed normally). This is fixed by clicking on the desktop, then try again.
Guess reason: I always use Super+Q (bound in "Keyboard" (dom0)) to close windows. Maybe the VM think it's stuck?

2.
When typing
```
> qvm-copy some-file
```
I press Enter to run the command. However, when the copy file dialog pops up, the VM *sometimes* thinks the Enter key is stuck.

Opening window in dom0 by pressing shortcuts registered in dom0 work.

Reason: The server side doesn't send any key event to unfocused window. This is a race condition between "release Enter" and "unfocus window".
Real reason: idk
Possible fix (haven't tried): Release all keys (to client) when window is really unfocused. To ignore jerking events from WM (rapidly: focusout -> focusin -> focusout), debounce the event by 1 second.

locriacyber:
I have no idea how to untangle this git history. I cleaned up the history up to the last merge commit.

gui-daemon/xside.c
```diff
@@ -374,10 +376,28 @@ static Window mkwindow(Ghandles * g, struct windowdata *vm_window)
...
+    xi_mask.mask_len = XIMaskLen(XI_LASTEVENT);
+    if (!(xi_mask.mask = calloc(xi_mask.mask_len, sizeof(char)))) {
+        fputs("Out of memory!\n", stderr);
+        exit(1);
+    }
+    XISetMask(xi_mask.mask, XI_KeyPress);
+    XISetMask(xi_mask.mask, XI_KeyRelease);
+    XISetMask(xi_mask.mask, XI_FocusIn);
+    XISetMask(xi_mask.mask, XI_FocusOut);
+    XISelectEvents(g->display, child_win, &xi_mask, 1);
```
DemiMarie, line 396: If this fails (returns anything but `Success`) the code should exit.

gui-daemon/xside.c
```diff
@@ -1870,6 +1915,16 @@ static void handle_configure_from_vm(Ghandles * g, struct windowdata *vm_window)
     }
 }
 
+static void send_keymap_notify(Ghandles * g)
+{
+    struct msg_hdr hdr;
+    char keys[32];
+    XQueryKeymap(g->display, keys);
```
DemiMarie, line 1922: If this does not succeed, just exit.

DemiMarie:
> I have no idea how to untangle this git history. I cleaned up the history up to the last merge commit.
> 
> Git is broken. I tried to do `git rebase -i --rebase-merges xxxxxxx` but I can't do `reword`, `fixup` or `squarch` easily. It should be automatic.

What do you normally use?  One option is to just create a patch series and use `git am` to apply it.  If the intermediate states would not be useful, just squash everything.

DemiMarie:
How is this normally handled by X applications?  That is, what event sequence do X applications normally get if a key is pressed while they have keyboard focus, and is released after they have lost focus?  Qubes should send the same event sequence in the same order.  This should fix other bugs as well.

locriacyber:
> How is this normally handled by X applications? That is, what event sequence do X applications normally get if a key is pressed while they have keyboard focus, and is released after they have lost focus? Qubes should send the same event sequence in the same order. This should fix other bugs as well.

We discussed about this earlier. The problem is
1. You can't send all events with X API. Some applications ignore `SendEvent` (I tried, maybe I did it wrong).
2. If focus being stolen by another VM, we can't "same event sequence in the same order". The events are spliced in time.

I suspect it's sometimes `B:FocusIn A:FocusOut` sometimes `A:FocusOut B:FocusIn`. It's not in X's spec (WM-implementation).

locriacyber:
Still one more bug that I can't fix:

When 1 VM switch from vm0:window0 to vm0:window1, I can't regain focus by sending focus events. This bug didn't exist before this PR.

Sending focus on mouse down is not working.

DemiMarie:
> When 1 VM switch from vm0:window0 to vm0:window1, I can't regain focus by sending focus events. This bug didn't exist before this PR.

How does the dom0 window manager give focus to a window?  It should be possible for the GUI agent to do the same with its windows.

> 1. You can't send all events with X API. Some applications ignore `SendEvent` (I tried, maybe I did it wrong).

Those events should be able to be sent via the X11 driver, which is part of the GUI agent.

locriacyber:
"Focus window when clicked" was originally handled by `vmside.c`. I removed it during refactoring. I'll add it back now properly.

> Those events should be able to be sent via the X11 driver, which is part of the GUI agent.
I meant some focus events are traditionally sent by like XFCE4, and Xlib API cannot send that.

DemiMarie:
> > Those events should be able to be sent via the X11 driver, which is part of the GUI agent.
> 
> Some special focus events are traditionally sent by like XFCE4, and Xlib API cannot send that.

Presumably there is _some_ API that XFCE4 is using; Qubes can use the same.

locriacyber:
I think XFCE4 use xf86 (library for window manager/ X driver)

gui-daemon/xside.c
DemiMarie: #104

DemiMarie:
@locriacyber Can you drop the focus-related changes from this PR?  I based #104 on them.

locriacyber:
I wrote them at the same time, and it would be difficult to separate in git history. Do you want just the focus part as separate branch?

Also, just having `xside.c` won't work well unless `vmside.c` sends the right `detail` and `mode` in focus events.

locriacyber:
The focus code has a bug: when clicking on IBus's tray icon, I get three simultaneous popups. I don't know if IBus think it's been clicked thrice, or it's something else.
Every other tray icon works fine.

locriacyber:
Is it possible to dynamically negotiate protocol?

Currently in vmside.c:

```
#if !(PROTOCOL_VERSION_MAJOR == QUBES_GUID_PROTOCOL_VERSION_MAJOR && \
      PROTOCOL_VERSION_MINOR <= QUBES_GUID_PROTOCOL_VERSION_MINOR)
#  error Incompatible qubes-gui-protocol.h.
#endif
```

it's hardcoded.

DemiMarie:
@locriacyber Not yet.  It will be starting with protocol 1.4 (#105).

DemiMarie:
@locriacyber Protocol 1.4 has been merged, so dynamic negotiation is now possible.

locriacyber:
This branch works, but go to #115 for the same features with backwards-compatibility.

PR #100, opened 2022-03-31
Clean up some macros

DemiMarie:
Add do-while loops and ensure that min() and max() only evaluate their
arguments once.


marmarek:
PipelineRetry

marmarek (changes requested):

marmarek (approved):

PR #104, opened 2022-07-03
Pass all focus events to clients

### LINKED ISSUE #7599: Spurious keystrokes while switching between windows
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release

R4.1, but R4.0 is probably affected too.

### Brief summary

When cycling between windows, qubes are sent keystroke events that they really should not be.

### Steps to reproduce

See https://github.com/QubesOS/qubes-gui-daemon/pull/98#discussion_r820048640 and related discussion.

### Expected behavior

No extraneous keystrokes.

### Actual behavior

Extraneous keystrokes.

DemiMarie:
Ignoring some focus events can result in the client and server
disagreeing on which window is focused or which keys have been pressed.
The results are unpredictable and generally undesirable.

Therefore, pass all focus events to clients.  The mode is still
overridden with NotifyNormal to avoid leaking window manager
implementation details.

Fixes QubesOS/qubes-issues#7599.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -1941,15 +1941,25 @@ static void process_xevent_focus(Ghandles * g, const XFocusChangeEvent * ev)
     struct msg_focus k;
     CHECK_NONMANAGED_WINDOW(g, ev->window);
 
-    /* Ignore everything other than normal, non-temporary focus change. In
-     * practice it ignores NotifyGrab and NotifyUngrab. VM does not have any
-     * way to grab focus in dom0, so it shouldn't care about those events. Grab
-     * is used by window managers during task switching (either classic task
-     * switcher, or KDE "present windows" feature).
+    /* In the past, the GUI daemon ignored several such events.  The
+     * comment used to say:
+     *
+     * > Ignore everything other than normal, non-temporary focus change. In
+     * > practice it ignores NotifyGrab and NotifyUngrab. VM does not have any
+     * > way to grab focus in dom0, so it shouldn't care about those events. Grab
+     * > is used by window managers during task switching (either classic task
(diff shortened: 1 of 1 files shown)


gui-daemon/xside.c
```diff
@@ -1959,8 +1968,8 @@ static void process_xevent_focus(Ghandles * g, const XFocusChangeEvent * ev)
     hdr.type = MSG_FOCUS;
     hdr.window = vm_window->remote_winid;
     k.type = ev->type;
-    /* override NotifyWhileGrabbed with NotifyNormal b/c VM shouldn't care
-     * about window manager details during focus switching
+    /* Override mode with NotifyNormal.  The VM shouldn't care
```
locriacyber, line 1971: Applications care very much though. Some depend of this information to decide where to ignore keystrokes or not (they want to ignore Alt-Tab when switching windows).

DemiMarie: Good point.

locriacyber: You still need client-side support. My hacky solution here:
https://github.com/QubesOS/qubes-gui-agent-linux/pull/154/files#diff-d2f30726ab5b690e7ac5631b2a30278386f40ada334a45b41787c59535b59333R1862

DemiMarie: Yikes.  This is ugly.

Can you include a separate PR to qubes-gui-agent-linux with just this change?

DemiMarie:
PipelineRetryFailed

DemiMarie:
PipelineRetryFailed

marmarek:
Linked issue is closed as not relevant on R4.1+ anymore

PR #105, opened 2022-07-06
Implement GUI protocol 1.4

DemiMarie:
This allows bidirectional protocol negotiation, which means that the
agent can be newer than the daemon.

The daemon must be upgraded to support protocol 1.4 first.  Afterwards,
agents and daemons can be independently upgraded.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -69,8 +69,8 @@
 /* Supported protocol version */
 
 #define PROTOCOL_VERSION_MAJOR 1
-#define PROTOCOL_VERSION_MINOR 3
-#define PROTOCOL_VERSION (PROTOCOL_VERSION_MAJOR << 16 | PROTOCOL_VERSION_MINOR)
+#define PROTOCOL_VERSION_MINOR 4
+#define PROTOCOL_VERSION(x, y) ((x) << 16 | (y))
 
 #if !(PROTOCOL_VERSION_MAJOR == QUBES_GUID_PROTOCOL_VERSION_MAJOR && \
       PROTOCOL_VERSION_MINOR <= QUBES_GUID_PROTOCOL_VERSION_MINOR)
@@ -1312,7 +1312,7 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
             if (len > 0) {
                 /* MSG_CLIPBOARD_DATA used to use the window field to pass the length
                    of the blob, be aware when working with old implementations. */
-                if (g->agent_version < 0x00010002)
+                if (g->protocol_version < PROTOCOL_VERSION(1, 2))
                     hdr.window = len;
                 else
                     hdr.window = remote_winid;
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -3596,16 +3596,32 @@ static void release_all_mapped_mfns(void)
...
+    return g->protocol_version >= ((uint32_t)major << 16 | (uint32_t)minor);
+}
+
 static void send_xconf(Ghandles * g)
 {
-    struct msg_xconf xconf;
+    struct {
+        uint32_t version;
+        struct msg_xconf xconf;
+    } conf;
```
marmarek, line 3609: There is no need to send those two messages at once, please use separate write (and no wrapping struct).

gui-daemon/xside.c
```diff
@@ -3596,16 +3596,32 @@ static void release_all_mapped_mfns(void)
...
-    xconf.mem = xconf.w * xconf.h * 4 / 1024 + 1;
-    write_struct(g->vchan, xconf);
+    if (XGetWindowAttributes(g->display, g->root_win, &attr) != Success)
+        err(1, "XGetWindowAttributes");
+    conf.version = g->protocol_version;
+    conf.xconf.w = _VIRTUALX(attr.width);
+    conf.xconf.h = attr.height;
+    conf.xconf.depth = attr.depth;
+    conf.xconf.mem = conf.xconf.w * conf.xconf.h * 4 / 1024 + 1;
+    if (protocol_version_at_least(g, 1, 4)) {
```
marmarek, line 3618: That's a lot of code for `g->protocol_version >= 0x00010004U` (it may be worth to have a macro for combining them, and maybe even a `#define` somewhere for this specific version value, as the point of changing negotiation).

marmarek (changes requested):

gui-daemon/xside.c
DemiMarie: Sending them at once is a teeny bit faster, but I admit that the loss in maintainability probably is not worthwhile.

gui-daemon/xside.c
DemiMarie: Are you referring to the `protocol_version_at_least` inline function?  To me, that is quite a bit clearer than `g->protocol_version >= 0x00010004U`, and it only needs to be written once.

marmarek: Yes, having a function to perform just `>=` hides what it actually is.
```
#define GUI_PROTOCOL(major, minor) ((major) << 16 | (minor))
...
if (g->protocol_version >= GUI_PROTOCOL_VERSION(1, 4)) {
```

DemiMarie:
Build is broken because this has a hard dependency on QubesOS/qubes-gui-common#16.

DemiMarie:
@locriacyber The problem with this is that it requires support from both daemon and agent, and right now neither have such support.  Therefore, there needs to be _some_ way to announce support for this.  Also, having to deal with all possible feature bits is a testing nightmare.  Just having to deal with all the different version combinations will be hard enough.

gui-daemon/xside.h
```diff
@@ -181,7 +181,7 @@ struct _global_handles {
     char vmname[32];    /* name of VM */
     int domid;        /* Xen domain id (GUI) */
     int target_domid;        /* Xen domain id (VM) - can differ from domid when GUI is stubdom */
-    uint32_t agent_version;  /* gui-agent (protocol) version */
+    uint32_t protocol_version;  /* gui-agent (protocol) version */
```
marmarek, line 184: Comment needs updating - it isn't just agent version anymore, it is negotiated protocol version (which may be older than the agent's one).

marmarek (changes requested):

marmarek (approved):

marmarek:
> Since newer clients will not work with older server.

Yes, but older clients will work with newer server.

PR #106, opened 2022-07-13
Release MFNs for unmapped windows as well

m-v-b:
Hi @marmarek and @DemiMarie,

I believe that I have found the root cause of the "GUI-related lingering Xen grant references" issue reported in https://github.com/QubesOS/qubes-issues/issues/7539, and I prepared this patch to resolve it.

As noted in the commit message, I have only tested this with Qubes OS v4.1, but I do not see why this commit would not work on the master branch.

Thank you,

Vefa

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -3503,6 +3503,7 @@ static void handle_message(Ghandles * g)
             break;
         vm_window->is_mapped = 0;
         (void) XUnmapWindow(g->display, vm_window->local_winid);
+        release_mapped_mfns(g, vm_window);
         break;
     case MSG_CONFIGURE:
         handle_configure_from_vm(g, vm_window);


m-v-b:
Whoops. I forgot to add a signature. Let's fix that.

DemiMarie (approved):
Nice catch @m-v-b!  Thanks for the patch and the awesome commit message.

If you have the time, would you mind submitting the agent-side patch you used as well?  The kernel patch would also be useful; I suggest submitting it to xen-devel@lists.xenproject.org.

marmarek:
An unmapped window is still a window and can be mapped again. gui-agent will try to send `MSG_WINDOW_DUMP` after mapping (at first damage event), but it isn't clear to me if that's guaranteed to happen. Have you tested various cases of unmapped and mapped again windows (common example are menus)?

As for the kernel debug, there are already some useful debug prints that can be enabled via `/sys/kernel/debug/dynamic_debug/control` - including printing deferred grant refs before timeout expires.

m-v-b:
Hi,

Thank you both for the feedback!

@DemiMarie, I can provide the patch for the GUI agent as well if it would be beneficial, but please note that it is very basic. The patch involves adding a loop to `send_pixmap_grant_refs` that iterates over the grant references and prints them out one by one. We would likely need a new command line argument to optionally enable this behaviour (or maybe a log_level > 2).

I doubt that the kernel changes would be accepted upstream, as modifying the time-out duration is very debugging oriented even though the patch itself does not break any rules.

As an aside, there is a minor unsigned integer wraparound on the following line that I noticed and locally fixed while debugging this issue: https://elixir.bootlin.com/linux/v5.15.54/source/drivers/xen/grant-table.c#L382

@marmarek, if I may answer your questions:

In terms of testing: Yes, I have used a variant of this approach on my personal laptop (with Qubes OS v4.1) over the past two days or so, and I did not notice regressions during my regular use of my laptop. 

Note that this initial approach involved modifying the GUI agent so that it calls `send_pixmap_grant_refs` before sending the `MSG_UNMAP` message. (My initial approach did not change the GUI daemon at all.) However, I later thought that carrying out the grant release operation in the GUI daemon (i.e., in dom0) would probably be safer, so that dom0 would not rely on code running in the VMs for the grant release operation. This prompted me to change the taken approach to the one in this pull request.

I also tested this specific patch as follows:

1) Use context and toolbar menus in various applications with different toolkits in a Fedora 36-based AppVM: Firefox (GTK3), VLC (Qt5), speedcrunch (Qt5), xterm (Xt), xfce4-terminal (GTK3).

2) Use the following xdotool command line on an xfce4-terminal instance while it is running the loop: `while true; do date -In ; sleep 0.5; done` . Here is the xdotool command line: `win=$(xdotool selectwindow); echo $win; xdotool windowunmap $win; sleep 10 ; xdotool windowmap $win` . (xdotool is run from another terminal in the same AppVM.)

In both tests, I did not notice any user-interface issues.

In terms of kernel debugging facilities: I agree that there are `pr_debug` calls that print out grant entries just before deferring their release operations. Note that I was interested in finding out which ones were **still** pending a release operation after a time-out, but faster. The default of 60 seconds was not very suitable for rapid testing, so I lowered it to a more manageable duration (e.g., 5~10 seconds) using a kernel parameter.

----
Thanks again for the feedback, Demi and Marek! I hope that my comment addresses your questions and concerns.

marmarek:
Thanks, that explains my concerns. I did some light testing myself and also haven't found any case where gui-agent wouldn't send MSG_WINDOW_DUMP just after mapping. I scheduled this PR for the next openQA run, just to be sure nothing else was missed (it does include some tests for Windows gui agent and also the one in stubdomain).

DemiMarie:
> As an aside, there is a minor unsigned integer wraparound on the following line that I noticed and locally fixed while debugging this issue: https://elixir.bootlin.com/linux/v5.15.54/source/drivers/xen/grant-table.c#L382

Would you mind sending this patch upstream?

m-v-b:
Hi @marmarek and @DemiMarie,

Just a heads-up that we might want to wait before merging this commit.

I had an incident where Firefox 102.0.1 misbehaved. Its window contents would not update even if I tried to resize it or attempt to interact with it. (The misbehaviour was similar to the "full screen" toggle issue from years ago.)

I will continue using this patch as normal to see if I reproduce the same issue. If I do, I will revert my patched gui-daemon package and attempt to reproduce it to understand the root cause. Due to the intermittent nature of the misbehaviour, this might take some time. I will keep you posted via comments in this pull request.

----

Hi @DemiMarie,

> > As an aside, there is a minor unsigned integer wraparound on the following line that I noticed and locally fixed while debugging this issue: https://elixir.bootlin.com/linux/v5.15.54/source/drivers/xen/grant-table.c#L382
> 
> Would you mind sending this patch upstream?

Ah, my apologies; I cannot make a promise, but I will see what I can do.

> > I doubt that the kernel changes would be accepted upstream, as modifying the time-out duration is very debugging oriented even though the patch itself does not break any rules.
> 
> I believe there are quite a few debug-only kernel parameters, so adding another tunable would be reasonable.

If I end up sending a patch for the aforementioned integer wraparound, I can send the debugging-oriented patch in the same time frame.

DemiMarie:
Thanks for being cautions @m-v-b!

marmarek:
> I had an incident where Firefox 102.0.1 misbehaved. Its window contents would not update even if I tried to resize it or attempt to interact with it. (The misbehaviour was similar to the "full screen" toggle issue from years ago.)

I hit this issue occasionally too, without this change. When it happens, you can roll up the window and then back again (mouse wheel on the title bar in xfce) to force redraw. If it successfully update the window content, it's "just" damage events delivery, unrelated to this PR.

m-v-b:
Hi again!

After a few hours of normal use of my laptop, I have just reproduced the aforementioned misbehaviour with Firefox 102.0.1 one more time, again with this pull request. Marek's description of the issue is consistent with what I observe now. Rolling the window up and down using the XFCE window manager refreshes the window's contents, and contrary to my original report, resizing the window also refreshes the window's contents.

Given that this appears to be a known issue, I do not have any further concerns regarding this pull request.

Thanks again for the feedback!

marmarek:
> I had been independently debugging the issue fixed by Demi, unaware of Demi's patch: https://lore.kernel.org/all/20220710230522.1563-1-demi@invisiblethingslab.com/

The patch just fixes excessive `WARN_ON`, it doesn't affect result of the mapping call. And since forcing redraw actually works, the issue is unlikely to be caused by any mapping/unmapping failure.

m-v-b:
> > I had been independently debugging the issue fixed by Demi, unaware of Demi's patch: https://lore.kernel.org/all/20220710230522.1563-1-demi@invisiblethingslab.com/
> 
> The patch just fixes excessive `WARN_ON`, it doesn't affect result of the mapping call.

If gntdev_mmap fails mid-way, then, as part of its clean-up, it will trigger the call to __unmap_grant_pages_done, which has the WARN_ON statements in question. (I had confirmed this with a few additional pr_debug calls.)

My additional pr_err messages in the __unmap_grant_pages_done function were triggered around the time frame of the Firefox misbehaviour on my system. I thought that perhaps gntdev_mmap failed due to a reason I do not know yet, which triggered the Firefox misbehaviour and my pr_err messages.

> And since forcing redraw actually works, the issue is unlikely to be caused by any mapping/unmapping failure.

Good point; I did not think of this.

marmarek (approved):

m-v-b:
Hi @marmarek and @DemiMarie,

Since having published this pull request, I have found that this pull request resolves only resolves the pending grant requests issue for a specific condition (unmapped windows).

However, there appears to be a deeper problem than only having lingering unmapped windows. When rapidly resizing VMs' GUI windows, I found that shmoverride library's grant mapping operations fail due to a failing "mmap" syscall made by the libxl libraries, with an errno code of 22. I traced this all the way to Xen (via code changes/print-outs in shmoverride, Linux and Xen), but have not been able to find a root cause, at least yet.

Adding to the complications, when Xen is built with debugging (CONFIG_DEBUG) enabled, resizing VM windows causes Xen to forcibly crash dom0 after Xen prints out the following error message: `(XEN) mm.c:1206:d0v0 Attempt to implicitly unmap a granted PTE ...`. Here is the call trace I acquired from this by patching Xen:

```
(XEN) Xen call trace:
(XEN)    [<ffff82d040206fac>] R __domain_crash+0xa6/0xaa
(XEN)    [<ffff82d0402fc522>] F put_page_from_l1e+0x129/0x13e
(XEN)    [<ffff82d0402ddb79>] F create_grant_pv_mapping+0x1ad/0x3a1
(XEN)    [<ffff82d040211bcd>] F grant_table.c#gnttab_map_grant_ref+0xa8c/0x1339
(XEN)    [<ffff82d040213532>] F do_grant_table_op+0xdf/0x1e49
(XEN)    [<ffff82d0402de531>] F pv_hypercall+0x278/0x2b9
(XEN)    [<ffff82d04035b4a9>] F lstar_enter+0x149/0x150
```

When I patch Xen to not crash dom0 like that (after all, Xen does not crash dom0 when CONFIG_DEBUG is disabled), I notice a flood of `(XEN) grant_table.c:772:d0v0 Bad flags (0) or dom (0); expected d0` when GUI windows are rapidly resized, and these logs coincide with the failed mmap calls made by shmoverride via libxl. I believe that this occurs because the page for which a grant entry was made had been mapped from the AppVM.

I also rolled back to Xen 4.14.0 and Linux 5.10.74, but the problems remained with those versions as well, so my guess is that this is a day-1 issue with the grant-based GUI protocol. (Note that I have been testing with Linux 5.15.55 with @DemiMarie's warning fix-up patch, along with my debugging patches.)

All this to say, I do not have any issues with this pull request being merged, but we might have a deeper problem. My current **guess** is that there may be a need to synchronize the operations between shmoverride and the AppVM's unsharing of granted pages to avoid these kinds of issues. Perhaps the AppVM should only unshare granted pages when it learns from dom0 that shmoverride is no longer making use of them.

marmarek:
> When rapidly resizing VMs' GUI windows, I found that shmoverride library's grant mapping operations fail due to a failing "mmap" syscall made by the libxl libraries, with an errno code of 22.

There is a legitimate case when this may happen: when a window composition buffer on the VM's side is invalidated before dom0 even manage to map it. Usual case is window that is destroyed quickly after being created, but resizing also may trigger this case. The old (pre-grant tables) version had this issue too, but the visible effect was different: instead of failed mapping, you got "just" garbage content in the window (as the old approach hasn't verified if the mapped page is still what it was supposed to be).

DemiMarie:
> Hi @marmarek and @DemiMarie,
> 
> Since having published this pull request, I have found that this pull request resolves only resolves the pending grant requests issue for a specific condition (unmapped windows).
> 
> However, there appears to be a deeper problem than only having lingering unmapped windows. When rapidly resizing VMs' GUI windows, I found that shmoverride library's grant mapping operations fail due to a failing "mmap" syscall made by the libxl libraries, with an errno code of 22. I traced this all the way to Xen (via code changes/print-outs in shmoverride, Linux and Xen), but have not been able to find a root cause, at least yet.

This is _probably_ harmless, but I need to make sure it cannot cause anything more serious.

> Adding to the complications, when Xen is built with debugging (CONFIG_DEBUG) enabled, resizing VM windows causes Xen to forcibly crash dom0 after Xen prints out the following error message: `(XEN) mm.c:1206:d0v0 Attempt to implicitly unmap a granted PTE ...`. Here is the call trace I acquired from this by patching Xen:
> 
> ```
> (XEN) Xen call trace:
> (XEN)    [<ffff82d040206fac>] R __domain_crash+0xa6/0xaa
> (XEN)    [<ffff82d0402fc522>] F put_page_from_l1e+0x129/0x13e
> (XEN)    [<ffff82d0402ddb79>] F create_grant_pv_mapping+0x1ad/0x3a1
> (XEN)    [<ffff82d040211bcd>] F grant_table.c#gnttab_map_grant_ref+0xa8c/0x1339
> (XEN)    [<ffff82d040213532>] F do_grant_table_op+0xdf/0x1e49
> (XEN)    [<ffff82d0402de531>] F pv_hypercall+0x278/0x2b9
> (XEN)    [<ffff82d04035b4a9>] F lstar_enter+0x149/0x150
> ```
> 
> When I patch Xen to not crash dom0 like that (after all, Xen does not crash dom0 when CONFIG_DEBUG is disabled), I notice a flood of `(XEN) grant_table.c:772:d0v0 Bad flags (0) or dom (0); expected d0` when GUI windows are rapidly resized, and these logs coincide with the failed mmap calls made by shmoverride via libxl. I believe that this occurs because the page for which a grant entry was made had been mapped from the AppVM.

This is a bug in either Xen or the dom0 kernel.  No amount of misuse of `/dev/xen/gntdev` should be able to make dom0 crash like that, even in debug builds of Xen.  Either Xen’s call to `domain_crash()` is incorrect, dom0 is using the hypercall interface incorrectly, or both.

> I also rolled back to Xen 4.14.0 and Linux 5.10.74, but the problems remained with those versions as well, so my guess is that this is a day-1 issue with the grant-based GUI protocol. (Note that I have been testing with Linux 5.15.55 with @DemiMarie's warning fix-up patch, along with my debugging patches.)
> 
> All this to say, I do not have any issues with this pull request being merged, but we might have a deeper problem. My current **guess** is that there may be a need to synchronize the operations between shmoverride and the AppVM's unsharing of granted pages to avoid these kinds of issues. Perhaps the AppVM should only unshare granted pages when it learns from dom0 that shmoverride is no longer making use of them.

This may be necessary for correctness, but it is not necessary to prevent dom0 from crashing.  dom0 should not crash no matter what userspace or other VMs do with grant tables.

andyhhp:
First, take this patch in Xen:
```diff
diff --git a/xen/arch/x86/mm.c b/xen/arch/x86/mm.c
index 5b81d5fbdbb2..b3393385ffb6 100644
--- a/xen/arch/x86/mm.c
+++ b/xen/arch/x86/mm.c
@@ -1232,7 +1232,7 @@ void put_page_from_l1e(l1_pgentry_t l1e, struct domain *l1e_owner)
         gdprintk(XENLOG_WARNING,
                  "Attempt to implicitly unmap a granted PTE %" PRIpte "\n",
                  l1e_get_intpte(l1e));
-        domain_crash(l1e_owner);
+        pv_inject_hw_exception(TRAP_gp_fault, 0);
     }
 #endif
 
```
and read the comment while applying it.

This is a bug in dom0, and Xen's behaviour in this case (in debug builds only) is deliberate.

m-v-b:
Hi @andyhhp, thank you for taking a look and offering a suggestion! I can try that. I hope that it would be okay to continue the discussion in the following new bug report that @DemiMarie opened: https://github.com/QubesOS/qubes-issues/issues/7631

PR #109, opened 2022-07-15
Fix pacat-simple-vchan

DemiMarie:
This fixes multiple bugs in pacat-simple-vchan.  See individual commit messages for details.

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -736,7 +736,7 @@ static void control_socket_callback(pa_mainloop_api *UNUSED(a),
         command_len += ret;
         if (ret == 0)
             break;
-    } while (!memchr(command_buffer, '\n', command_len));
+    } while (!memchr(command_buffer + (command_len-ret), '\n', ret));
 
     if (strncmp(command_buffer, "audio-input 0\n", command_len) == 0) {
         new_rec_allowed = 0;
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -195,27 +195,34 @@ static void start_drain(struct userdata *u, pa_stream *s) {
 
 static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_length)
 {
-    int l = 0;
-    size_t index, buffer_length;
+    int l = 0, space_in_vchan;
+    size_t buffer_length;
     void *buffer = NULL;
     assert(s);
 
-    if (!libvchan_data_ready(u->play_ctrl) || !max_length)
+    if (!max_length)
         return;
 
-    buffer_length = max_length;
+    if ((space_in_vchan = libvchan_data_ready(u->play_ctrl)) < 0) {
+        pacat_log("libvchan_data_ready() failed: return value %d", space_in_vchan);
+        quit(u, 1);
+        return;
+    }
 
-    if (pa_stream_begin_write(s, &buffer, &buffer_length) < 0) {
+    buffer_length = (size_t)space_in_vchan > max_length ? max_length : (size_t)space_in_vchan;
+    if (!buffer_length)
+        return;
+
+    if (pa_stream_begin_write(s, &buffer, &buffer_length) || !buffer) {
         pacat_log("pa_stream_begin_write() failed: %s", pa_strerror(pa_context_errno(u->context)));
         quit(u, 1);
         return;
     }
-    index = 0;
 
-    while (libvchan_data_ready(u->play_ctrl) && buffer_length > 0) {
-        l = libvchan_read(u->play_ctrl, buffer + index, buffer_length);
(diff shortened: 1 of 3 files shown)


DemiMarie:
PipelineRefresh

DemiMarie:
I tested this, and it works.  However, when used with the standard (PulseAudio) agent, it introduces glitches that were not present previously.  I believe the culprit is the standard agent’s small buffer sizes, which lead to overruns.  The PipeWire agent avoids this by using larger buffers.

marmarek (changes requested):

pulse/pacat-simple-vchan.c
```diff
@@ -266,18 +276,36 @@ static void send_rec_data(pa_stream *s, struct userdata *u) {
...
-        if ((l=libvchan_write(u->rec_ctrl, rec_buffer + rec_buffer_index, rec_buffer_length)) < 0) {
-            pacat_log("libvchan_write failed");
-            quit(u, 1);
+    if ((vchan_buffer_space = libvchan_buffer_space(u->rec_ctrl)) < 0) {
+        pacat_log("libvchan_buffer_space failed");
+        quit(u, 1);
+        return;
+    }
+
+    if (rec_buffer_length > (size_t)vchan_buffer_space) {
```
marmarek, line 288: Is it guaranteed to _not_ trigger in normal usage (empty or almost-empty buffer)? IOW, is pulseaudio aware to send data in chunks smaller than 2048 bytes?

DemiMarie: With the current buffer sizes, this triggers quite frequently in my tests.  That said, the intent is to solve this by increasing the vchan buffer size (QubesOS/qubes-gui-agent-linux#164).  Should I handle it instead by buffering in this program?

DemiMarie: Good news: with a buffer size of 1 << 15 I get no overruns.

pulse/pacat-simple-vchan.c
```diff
@@ -266,18 +276,36 @@ static void send_rec_data(pa_stream *s, struct userdata *u) {
...
-        }
-        rec_buffer_length -= l;
-        rec_buffer_index += l;
+        size_t bytes_to_discard = rec_buffer_length - (size_t)vchan_buffer_space;
+        pacat_log("Overrun: discarding %zu bytes", bytes_to_discard);
+        rec_buffer += bytes_to_discard;
+        rec_buffer_length = (size_t)vchan_buffer_space;
     }
+
+    if ((l=libvchan_send(u->rec_ctrl, rec_buffer, rec_buffer_length)) < 0) {
```
marmarek, line 297: It should remain `libvchan_write`, otherwise you still break the old version (if vchan buffer is smaller than `rec_buffer_length`, it will never send anything). Furthermore, if `!u->never_block`, I think it should still loop over the whole buffer, not just attempt once.

marmarek (changes requested):

marmarek:
All recording tests failed with "only silence detected, no useful audio data". I guess that's because of the change I commented above.

DemiMarie:
New version pushed, but not tested.

DemiMarie:
Works for me with the stock PulseAudio module.

pulse/pacat-simple-vchan.c
```diff
@@ -266,12 +276,30 @@ static void send_rec_data(pa_stream *s, struct userdata *u) {
...
 
-    while (rec_buffer_length > 0 && u->rec_allowed) {
+    if (u->never_block) {
+        if ((vchan_buffer_space = libvchan_buffer_space(u->rec_ctrl)) < 0) {
+            pacat_log("libvchan_buffer_space failed");
+            quit(u, 1);
+            return;
+        }
+
+        if (u->never_block && rec_buffer_length > (size_t)vchan_buffer_space) {
```
marmarek, line 289: `u->never_block` is always true in this branch

pulse/pacat-simple-vchan.c
```diff
@@ -266,12 +276,30 @@ static void send_rec_data(pa_stream *s, struct userdata *u) {
...
+                return;
+            size_t bytes_to_discard = rec_buffer_length - (size_t)vchan_buffer_space;
+            pacat_log("Overrun: discarding %zu bytes", bytes_to_discard);
+            rec_buffer += bytes_to_discard;
+            rec_buffer_length = (size_t)vchan_buffer_space;
+        }
+    }
+
+    while (rec_buffer_length > 0) {
         /* can block */
```
marmarek, line 300: update comment?

marmarek (changes requested):

DemiMarie:
PipelineRetryFailed

locriacyber:
this unnecessarily depend on glib-2.0

marmarek (approved):

PR #110, opened 2022-07-22
xside: do not release buffer of the "whole screen" window

marmarek:
That window is backing buffers of other windows too. Windows GUI agent
depends on this feature. The window itself is mapped/unmapped when
switching between seamless and fill desktop modes.

Fixes QubesOS/qubes-issues#7643


DemiMarie (commented):
Looks good!

DemiMarie:
Also, this likely means we need more testing for Windows qubes.

PR #111, opened 2022-07-28
Support pipewire-pulseaudio as well as pulseaudio

DemiMarie:
pacat-simple-vchan works with PipeWire's PulseAudio server, not just
with PulseAudio itself.  Depend on the pulseaudio-daemon feature, which
is provided by both pipewire-pulesaudio and pulseaudio.


DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #117, opened 2022-08-30
Fix function prototype mismatches and use poll(2)

DemiMarie:
This fixes function prototype mismatches (caught using
-Wmissing-prototypes) and replaces select(2) with poll(2).  No change in
behavior.

### DIFF
--- a/gui-common/txrx-vchan.c
+++ b/gui-common/txrx-vchan.c
@@ -23,10 +23,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <libvchan.h>
-#include <sys/select.h>
 #include <errno.h>
-#include "double-buffer.h"
+#include <poll.h>
 #include <assert.h>
+
+#include <double-buffer.h>
+#include "../include/txrx.h"
 
 void (*vchan_at_eof)(void) = NULL;
 int vchan_is_closed = 0;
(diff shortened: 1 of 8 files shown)


gui-common/txrx-vchan.c
```diff
@@ -107,31 +109,19 @@ int read_data(libvchan_t *vchan, char *buf, int size)
...
-    FD_SET(vfd, &rfds);
-    if (vfd > max)
-        max = vfd;
-    max++;
-    ret = select(max, &rfds, NULL, NULL, &tv);
-    if (ret < 0 && errno == EINTR)
-        return -1;
+    struct pollfd fds[] = {
+        { .fd = libvchan_fd_for_select(vchan), .events = POLLIN, .revents = 0 },
+        { .fd = fd, .events = POLLIN | POLLRDHUP | POLLPRI, .revents = 0 },
```
marmarek, line 118: This is change in behavior - the `select()` version put the fd only in `readfds` set, _not_ `exceptfds`. Please keep the old behavior (just `POLLIN`), unless there is some specific reason for the change.

marmarek (changes requested):

marmarek:
Now this one needs rebase...

DemiMarie:
And rebased.

DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #121, opened 2022-09-30
Stricter command line parsing

DemiMarie:
See individual commit messages for details.

PR #122, opened 2022-10-03
Fix reexec if -- is given on the command line

DemiMarie:
The previous code appended the -f option.  This will not work if the command line contains -- to mark the end of options.  Instead, make -f argv[1], so that it always comes before the end-of-options marker.

Also add a do-while loop around a macro.


marmarek:
Unnecessary since #121. And also, `qvm-start-daemon` does not append `--`.

PR #123, opened 2022-10-15
do_shm_update: Negative width or height is an error

DemiMarie:
Negative width or height makes no sense.  Treat this as an error condition.  This allows simplifying the rest of do_shm_update(), as this corner case no longer needs to be checked for.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2011,7 +2011,7 @@ static void do_shm_update(Ghandles * g, struct windowdata *vm_window,
         // negative
 
         if (vm_window->y < 0 && vm_window->y+untrusted_y < 0) {
-            // we know vm_window->x is not less than -MAX_WINDOW_WIDTH, so this
+            // we know vm_window->y is not less than -MAX_WINDOW_HEIGHT, so this
             // is not UB.
             untrusted_y = -vm_window->y;
         }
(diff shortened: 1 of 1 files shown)


DemiMarie:
PipelineRetryFailed

gui-daemon/xside.c
```diff
@@ -1969,17 +1969,19 @@ static void process_xevent_focus(Ghandles * g, const XFocusChangeEvent * ev)
...
 static void do_shm_update(Ghandles * g, struct windowdata *vm_window,
-           int untrusted_x, int untrusted_y, int untrusted_w,
-           int untrusted_h)
+           int32_t untrusted_x, int32_t untrusted_y,
+           int32_t untrusted_w, int32_t untrusted_h)
 {
     int border_width = BORDER_WIDTH;
     int border_padding = 0; /* start forced border x pixels from the edge */
-    int x = 0, y = 0, w = 0, h = 0;
+    int32_t x = 0, y = 0, w = 0, h = 0;
```
marmarek, line 1977: I don't like this (and other in this PR) narrowing variables. It makes things more likely to explode if any of the reasoning below is incorrect (and also makes this reasoning more tricky, as for example adding two 32-bit values is no longer always-safe).

So, to this whole PR: refusing negative sizes early - okay, narrowing types - not.

marmarek (changes requested):

gui-daemon/xside.c
DemiMarie: I’m confused by what you mean by “narrowing types”.  `int32_t` is just a typedef for `int` on x86_64, so replacing `int` with `int32_t` is a no-op.  If you mean the `(int32_t)` casts, those are to suppress `-Wsign-conversion` and friends.  That said, those should be a separate PR.

marmarek (approved):

PR #125, opened 2022-10-29
Strictly validate message lengths

DemiMarie:
The future Wayland agent will include a proxy for the X11 agent.  This proxy will rely on the `untrusted_len` field for message framing.  If the value in this field does not match the number of bytes the GUI daemon actually reads, chaos will ensue.

To catch this problem before it can create havoc, validate that the `untrusted_len` field of incoming messages is correct.  Furthermore, any validation error that happens before all input has been read is now fatal, as continuing would leave data in the input stream that would be parsed as future protocol messages.  Therefore, do not use `VERIFY()`, which asks the user if they wish to continue.  Instead, log a fatal error and exit.

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -67,8 +67,8 @@
 
 /* Supported protocol version */
 
-#define PROTOCOL_VERSION_MAJOR 1u
-#define PROTOCOL_VERSION_MINOR 5u
+#define PROTOCOL_VERSION_MAJOR UINT32_C(1)
+#define PROTOCOL_VERSION_MINOR UINT32_C(6)
 #define PROTOCOL_VERSION(x, y) ((x) << 16 | (y))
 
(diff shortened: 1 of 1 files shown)


gui-daemon/xside.c
```diff
@@ -75,8 +75,8 @@
 
 /* Supported protocol version */
 
-#define PROTOCOL_VERSION_MAJOR 1u
-#define PROTOCOL_VERSION_MINOR 5u
+#define PROTOCOL_VERSION_MAJOR UINT32_C(1)
+#define PROTOCOL_VERSION_MINOR UINT32_C(6)
```
marmarek, line 79: We need a process for protocol changes, not just bumping version in various repositories.
So, lets do it this way: every protocol change (whether it's some new message, or "just" behavior change) first needs a PR to gui-common repo.
To be clear: it's okay to have an implementation PR open in parallel, but the merge order must be first gui-common with the formal protocol change and only then the implementation (which may require changing the version at that point - depending what happened in the meantime).

 I'll make a PR with initial content.

marmarek (approved):

PR #130, opened 2023-04-12
Drain playback stream before corking it

DemiMarie:
This ensures that there is no data remaining in the playback vchan. Otherwise, this data would be played when the VM _next_ starts playing audio, which is clearly wrong.

Reported-by: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>
Fixes: QubesOS/qubes-issues#8137

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -200,9 +200,6 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
     void *buffer = NULL;
     assert(s);
 
-    if (!max_length)
-        return;
-
     if ((space_in_vchan = libvchan_data_ready(u->play_ctrl)) < 0) {
         pacat_log("libvchan_data_ready() failed: return value %d", space_in_vchan);
         quit(u, 1);
@@ -211,7 +208,7 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
 
     buffer_length = (size_t)space_in_vchan > max_length ? max_length : (size_t)space_in_vchan;
     if (!buffer_length)
-        return;
+        goto maybe_cork;
 
     if (pa_stream_begin_write(s, &buffer, &buffer_length) || !buffer) {
         pacat_log("pa_stream_begin_write() failed: %s", pa_strerror(pa_context_errno(u->context)));
@@ -241,8 +238,16 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
             quit(u, 1);
             return;
         }
+        space_in_vchan -= (int)buffer_length;
     } else
         pa_stream_cancel_write(s);
+
+maybe_cork:
+    if (u->pending_play_cork && space_in_vchan == 0) {
+        pacat_log("Playback stream drained; corking it");
+        pa_stream_cork(u->play_stream, 1, NULL, u);
+        u->pending_play_cork = false;
+    }
 
     return;
 }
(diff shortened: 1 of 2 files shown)


marmarek:
This changes the behavior a bit, now a shorter part is played at next playback, but still some.
Reading [cork description](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a14e698233ac2d246646651955ab0ec7b), it seems to be intended for pausing intentionally _without_ draining the stream first ("as quickly as possible"). There is `pa_stream_drain()` that can be used for draining first.

I'm a bit worried about pause latency, but I think with the current buffer size it should be fine (if my calculations are correct, `0x8000` is about 186ms, so below 200ms I was talking about).

marmarek:
> `[Uncork playback stream if VM writes to it](/QubesOS/qubes-gui-daemon/pull/130/commits/8ce0cc931799deab1eb9975ff46f925d25f9945b) `

I'm not sure if this is a good idea, it may also uncork when it was not intended. What's exactly the issue? IIUC samples written to a corked stream will be played as soon as it gets uncorked, so it should be fine (as long as the uncork command is processed eventually), no?

marmarek:
If I guess your issue correctly, you need to set `pending_play_cork = false` when processing uncork command.

DemiMarie:
> If I guess your issue correctly, you need to set `pending_play_cork = false` when processing uncork command.

I figured that out myself :slightly_smiling_face: 

> > [Uncork playback stream if VM writes to it](/QubesOS/qubes-gui-daemon/pull/130/commits/8ce0cc931799deab1eb9975ff46f925d25f9945b)
> 
> I'm not sure if this is a good idea, it may also uncork when it was not intended. What's exactly the issue? IIUC samples written to a corked stream will be played as soon as it gets uncorked, so it should be fine (as long as the uncork command is processed eventually), no?

If the agent sends audio samples to a stream it does not intend to uncork immediately, I consider that to be an agent bug that should be fixed.  Much more likely is that the audio samples arrived before the uncork command did due to an unavoidable race.

marmarek:
> If the agent sends audio samples to a stream it does not intend to uncork immediately, I consider that to be an agent bug that should be fixed.

That is a likely option indeed, and should be fixed on the agent side. But implicit uncork based on data arriving on another channel may also cancel out pending cork. There may be also more complex race conditions, better not risk it even more.

pulse/pacat-simple-vchan.c
```diff
@@ -919,18 +996,25 @@ static void vchan_rec_async_connect(pa_mainloop_api *UNUSED(a),
     u->mainloop_api->io_free(u->rec_ctrl_event);
     u->rec_ctrl_event = NULL;
 
-    u->never_block = libvchan_buffer_space(u->rec_ctrl) > 8192;
```
marmarek, line 922: This was intentionally `>`, not `>=`, to not break pulseaudio agent that uses 8192 buffer size.

DemiMarie: Ah for some reason I thought it used 2048.

marmarek:
This seems to work now, but I see repeated `Playback stream drained; corking it` message (for pipewire agent - twice, for pulse agent, 6 times). I think that's misleading message, as it's printed before calling `playback_stream_drain()`, not before actual `pa_stream_cork()` in `playback_stream_drain_cb()`.

marmarek:
> * TC_20_AudioVM_PipeWire_fedora-37: [test_226_audio_playback_pipewire](https://openqa.qubes-os.org/tests/70821#step/TC_20_AudioVM_PipeWire_fedora-37/1) (failure)
>       `AssertionError: frequency 2073.4168195316165 not in specified range`

This is still happening, I'm not sure why. I think this wants saving recorded data to a file for manual inspection, but that can be a followup work after merge.

marmarek (approved):

PR #132, opened 2023-06-03
Allow periods in qube names

DemiMarie:
This allows periods in qube names by using the function from libqubes-pure to validate them instead of the built-in version.

Part of QubesOS/qubes-issues#8199.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,8 @@ Build-Depends:
  libpng-dev,
  libnotify-dev,
  qubesdb-dev,
- help2man
+ help2man,
+ libqubes-pure-dev,
 Standards-Version: 4.1.3
 Homepage: https://qubes-os.org/
 Vcs-Browser: https://github.com/QubesOS/qubes-gui-daemon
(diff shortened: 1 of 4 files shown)


marmarek:
PipelineRetry

debian/control
```diff
@@ -19,7 +19,8 @@ Build-Depends:
  libpng-dev,
  libnotify-dev,
  qubesdb-dev,
- help2man
+ help2man,
+ qubes-utils-dev (>= 4.2.10)
```
marmarek, line 23: There is no such package. It seems you put it into `libqubes-rpc-filecopy-dev`.

marmarek (changes requested):
This fails to build due to not linking with libqubes-pure.

debian/control
DemiMarie: Should I change that?  `libqubes-rpc-filecopy-dev` is obviously wrong.

marmarek: The proper Debian way would be a new subpackage. I think we don't link against that library anywhere yet (this is the first PR doing it, so it's still ok to fix packaging there.

DemiMarie: I’m pretty sure that qfile-unpacker links to libqubes-pure already.

marmarek: But fortunately, not directly, but through a library that is also in linux-utils, so splitting the package should still be safe-ish. If you want, I can do it, and try to setup dependencies properly.

DemiMarie: That would be great, thanks.

marmarek: Package is split, you can adjust dependency now.

marmarek (approved):

PR #133, opened 2023-09-17
Allow periods in qube names

DemiMarie:
This uses the validation function from libqubes-pure (which permits periods) instead of the old built-in version (which does not).

Part of QubesOS/qubes-issues#8199.

This reverts commit 1ba805168621ceaadf1610816d3299eb474473e8.


marmarek:
This fails to build (missing include?)

DemiMarie:
fixed

marmarek (approved):

PR #134, opened 2023-12-11
Fix window resizes under Wayland

DemiMarie:
If a configure event is received from the X server before the GUI agent acks a MSG_CONFIGURE, the ack from the GUI agent will be ignored.  This prevents the GUI agent from moving one edge of the window in response to the user moving the other edge.

Fixes: QubesOS/qubes-issues#8762

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -1720,6 +1720,16 @@ static void process_xevent_configure(Ghandles * g, const XConfigureEvent * ev)
             (int) vm_window->remote_winid, ev->width,
             ev->height, vm_window->width, vm_window->height,
             ev->x, ev->y, vm_window->x, vm_window->y);
+
+    if (ev->width < 1 || ev->height < 1 ||
+        ev->width > MAX_WINDOW_WIDTH || ev->height > MAX_WINDOW_HEIGHT) {
+        fprintf(stderr, "%s: got event from X server with bogus width/height (%dx%d)\n",
+                __func__, ev->width, ev->height);
+        return;
+    }
+
+    unsigned int width = (unsigned)ev->width, height = (unsigned)ev->height;
+
     /* non-synthetic events are about window position/size relative to the embeding
      * frame window (if applies), synthetic one (produced by window manager) are
      * about window position relative to original window parent.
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -1735,6 +1735,18 @@ static void process_xevent_configure(Ghandles * g, const XConfigureEvent * ev)
...
+    // If AppVM has not acknowledged the most recent configure message
+    // and this is not an override-redirect window, ignore this event.
+    // Otherwise, the AppVM's own ack would get ignored.  Do not do this
+    // for override-redirect windows because the daemon must always know
+    // the correct size and position of these windows.
+    //
+    // FIXME: there is still a race condition if the immediately following
+    // event is a resize of this window.
+    if (vm_window->have_queued_configure && !vm_window->override_redirect)
+        return;
```
marmarek, line 1748: There is similar return down below already. *If* this change is correct, the other one should be removed now, right?

But more generally, with this change, I think AppVM will miss configure messages if it isn't responding fast enough. Most importantly - will not get the message about final window position/size at all. Note the code in `handle_configure_from_vm` doesn't just ignore mismatching "ack" - it sends (or possibly re-sends) latest MSG_CONFIGURE, so if AppVM processes `MSG_CONFIGURE` and ack them as they arrive, it should arrive at the right coordinates finally (at last when user stops moving/resizing the window).

DemiMarie: This return only applies to managed windows, so it is not redundant with the other one.  That said, you are right that this change is buggy.

The problem I am trying to avoid is that when dragging the right edge of a non-resizeable Wayland surface, the compositor acks the `MSG_CONFIGURE` and then sends its own `MSG_CONFIGURE` with the size chosen by the Wayland client.  The subsequent `MSG_CONFIGURE` from the daemon is indistinguishable from user dragging the _left_ edge of the window and changing the right edge by 1 pixel.  This is clearly wrong.

To fix this, I think the GUI daemon needs to track which edges have been changed, and accept changes to edges that have not been moved in the meantime.

marmarek: What about telling the GUI daemon that a window shouldn't in fact be resized? You can do this with `MSG_WINDOW_HINTS`.

DemiMarie: I’ll need to see if the Wayland client makes that information available.  However, this is not the only case where moving one edge of a window can cause another edge to move.  Another is an image viewer that must preserve aspect ratio.

PR #138, opened 2024-01-25
Always block on vchan I/O by default

DemiMarie:
This significantly improves audio quality compared to nonblocking mode. I believe this is because the PipeWire agent will usually be able to free up space in the capture vchan quickly enough to avoid an overrun.

This has not been tested directly, but it is equivalent to always passing -b as a command-line argument, which I did test.

Fixes: QubesOS/qubes-issues#8887


marmarek:
@DemiMarie this alone seems to works just fine for both recording and playback, at least according to the automated tests
I'll do manual test with Firefox now

marmarek (approved):

PR #140, opened 2024-02-22
audio: change "low latency" buffer parameters

marmarek:
Related to QubesOS/qubes-issues#8955

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -215,6 +215,16 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
         return;
     }
 
+    if (verbose > 1) {
+        pa_usec_t latency = 0;
+        int negative;
+
+        if (pa_stream_get_latency(s, &latency, &negative))
+            pacat_log("pa_stream_get_latency() failed");
+
+        pacat_log("process_playback_data(): vchan data %d max_length %d latency %llu", space_in_vchan, max_length, latency);
+    }
+
     buffer_length = (size_t)space_in_vchan > max_length ? max_length : (size_t)space_in_vchan;
     if (!buffer_length)
         goto maybe_cork;
@@ -669,6 +679,8 @@ static void context_state_callback(pa_context *c, void *userdata) {
             pa_stream_set_event_callback(u->play_stream, stream_event_callback, u);
             pa_stream_set_buffer_attr_callback(u->play_stream, stream_buffer_attr_callback, u);
             flags = PA_STREAM_ADJUST_LATENCY;
+            if (verbose > 1)
+                flags |= PA_STREAM_AUTO_TIMING_UPDATE | PA_STREAM_INTERPOLATE_TIMING;
 
             if (pa_stream_connect_playback(u->play_stream, u->play_device, bufattr, flags, NULL /* volume */, NULL) < 0) {
                 pacat_log("pa_stream_connect_playback() failed: %s", pa_strerror(pa_context_errno(c)));
@@ -1025,6 +1037,7 @@ static _Noreturn void usage(char *arg0, int arg) {
     fprintf(stream, "  -l - low-latency mode (higher CPU usage)\n");
     fprintf(stream, "  -n - never block on vchan I/O (overrides previous -b option)\n");
     fprintf(stream, "  -b - always block on vchan I/O (default, overrides previous -n option)\n");
+    fprintf(stream, "  -v - verbose logging (a lot of output, may affect performance)\n");
     fprintf(stream, "  -h - print this message\n");
     if (fflush(NULL) || ferror(stdout) || ferror(stderr))
         exit(1);
@@ -1046,7 +1059,7 @@ int main(int argc, char *argv[])
     memset(&u, 0, sizeof(u));
     if (argc <= 2)
         usage(argv[0], 1);
-    while ((i = getopt(argc, argv, "+lnbh")) != -1) {
+    while ((i = getopt(argc, argv, "+lnbvh")) != -1) {
         switch (i) {
             case 'l':
                 bufattr = &custom_bufattr;
@@ -1056,6 +1069,9 @@ int main(int argc, char *argv[])
                 break;
             case 'b':
                 u.never_block = false;
+                break;
+            case 'v':
+                verbose += 1;
                 break;
             case 'h':
                 usage(argv[0], 0);
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -100,6 +100,11 @@ static void playback_stream_drain_cb(pa_stream *s, int success, void *userdata);
 
 void pacat_log(const char *fmt, ...) {
     va_list args;
+    struct timespec ts;
+
+    clock_gettime(CLOCK_REALTIME, &ts);
+
+    fprintf(stderr, "%10jd.%06ld ", ts.tv_sec, ts.tv_nsec / 1000);
     va_start(args, fmt);
     vfprintf(stderr, fmt, args);
     va_end(args);
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -81,11 +81,11 @@ static pa_sample_spec sample_spec = {
     .rate = 44100,
     .channels = 2
 };
-static const pa_buffer_attr custom_bufattr ={
+static pa_buffer_attr custom_bufattr ={
     .maxlength = (uint32_t)-1,
     .minreq = (uint32_t)-1,
     .prebuf = (uint32_t)-1,
-    .fragsize = 4096,
+    .fragsize = 2048,
     .tlength = 4096
 };
 const pa_buffer_attr * bufattr = NULL;
@@ -1032,12 +1032,14 @@ static void vchan_rec_async_connect(pa_mainloop_api *UNUSED(a),
 
 static _Noreturn void usage(char *arg0, int arg) {
     FILE *stream = arg ? stderr : stdout;
-    fprintf(stream, "usage: %s [-l] [--] domid domname\n",
+    fprintf(stream, "usage: %s [options] [--] domid domname\n",
             arg0 ? arg0 : "pacat-simple-vchan");
     fprintf(stream, "  -l - low-latency mode (higher CPU usage)\n");
     fprintf(stream, "  -n - never block on vchan I/O (overrides previous -b option)\n");
     fprintf(stream, "  -b - always block on vchan I/O (default, overrides previous -n option)\n");
     fprintf(stream, "  -v - verbose logging (a lot of output, may affect performance)\n");
+    fprintf(stream, "  -t size - target playback buffer fill, implies -l, default %d\n",
+            custom_bufattr.tlength);
     fprintf(stream, "  -h - print this message\n");
     if (fflush(NULL) || ferror(stdout) || ferror(stderr))
         exit(1);
@@ -1055,11 +1057,13 @@ int main(int argc, char *argv[])
     int pidfile_fd;
     int play_watch_fd, rec_watch_fd;
     int i;
+    unsigned long tlength;
+    char *endptr;
 
     memset(&u, 0, sizeof(u));
     if (argc <= 2)
         usage(argv[0], 1);
-    while ((i = getopt(argc, argv, "+lnbvh")) != -1) {
+    while ((i = getopt(argc, argv, "+lnbvt:h")) != -1) {
         switch (i) {
             case 'l':
                 bufattr = &custom_bufattr;
@@ -1073,6 +1077,18 @@ int main(int argc, char *argv[])
             case 'v':
                 verbose += 1;
                 break;
+            case 't':
+                errno = 0;
+                tlength = strtoul(optarg, &endptr, 0);
+                if (*endptr)
+                    errx(1, "Invalid -t argument: %s", optarg);
+                if (tlength > UINT32_MAX)
+                    errno = ERANGE;
+                if (errno)
+                    err(1, "Invalid -t argument: %s", optarg);
+                bufattr = &custom_bufattr;
+                custom_bufattr.tlength = tlength;
+                break;
             case 'h':
                 usage(argv[0], 0);
             default:
@@ -1082,7 +1098,6 @@ int main(int argc, char *argv[])
     if (argc - optind != 2)
         usage(argv[0], 1);
     const char *domid_str = argv[optind], *domname = argv[optind + 1];
-    char *endptr;
     errno = 0;
     long l_domid = strtol(domid_str, &endptr, 10);
     /* 0x7FF0 is DOMID_FIRST_RESERVED */


marmarek:
FYI @DemiMarie I've added logging I used for graphs here. Since it proved useful, I'll make it opt-in in default binary too.

pulse/pacat-simple-vchan.c
```diff
@@ -82,11 +82,11 @@ static pa_sample_spec sample_spec = {
     .channels = 2
 };
 static const pa_buffer_attr custom_bufattr ={
-    .maxlength = (uint32_t)-1,
+    .maxlength = 8192,
     .minreq = (uint32_t)-1,
     .prebuf = (uint32_t)-1,
     .fragsize = 4096,
-    .tlength = 4096
+    .tlength = (uint32_t)-1
```
DemiMarie, line 89: Is this reasonable?  The documentation says:

```c
    /**< Playback only: target length of the buffer. The server tries                         
     * to assure that at least tlength bytes are always available in                          
     * the per-stream server-side playback buffer. The server will                            
     * only send requests for more data as long as the buffer has                             
     * less than this number of bytes of data.                                                
     *
     * It is recommended to set this to (uint32_t) -1, which will                             
     * initialize this to a value that is deemed sensible by the
     * server. However, this value will default to something like 2s;                         
     * for applications that have specific latency requirements                               
     * this value should be set to the maximum latency that the                               
     * application can deal with.                                                             
     *
     * When PA_STREAM_ADJUST_LATENCY is not set this value will                               
     * influence only the per-stream playback buffer size. When
     * PA_STREAM_ADJUST_LATENCY is set the overall latency of the sink                        
     * plus the playback buffer size is configured to this value. Set
     * PA_STREAM_ADJUST_LATENCY if you are interested in adjusting the                        
     * overall latency. Don't set it if you are interested in                                 
     * configuring the server-side per-stream playback buffer                                 
     * size. */                                                                               
```

2s is way, *way* too high.  I would much prefer to keep this at 4096.

pulse/pacat-simple-vchan.c
```diff
@@ -82,11 +82,11 @@ static pa_sample_spec sample_spec = {
     .channels = 2
 };
 static const pa_buffer_attr custom_bufattr ={
-    .maxlength = (uint32_t)-1,
+    .maxlength = 8192,
```
DemiMarie, line 85:
```suggestion
    .maxlength = 4096,
```

DemiMarie (changes requested):
I am quite concerned about a latency regression here.  I would prefer to set `.maxlength` to 4096, while keeping `.tlength` at 4096.

pulse/pacat-simple-vchan.c
marmarek: It can't be higher than maxlength (which is buffer size).

pulse/pacat-simple-vchan.c
marmarek: No. `PA_STREAM_ADJUST_LATENCY` is enabled, so this includes whatever buffer hardware has (which in fact may be even larger than 8k). Furthermore, PW agent sends data in 4096 chunks, so having maxlength 4k means missing just one write will lead to underrun.

marmarek:
> I am quite concerned about a latency regression here

It isn't regression at all. maxlength used to be server default, which IIUC is about 4MB. PA agent used smaller vchan buffer, so it didn't manage to fill the whole PA's buffer, but it was exactly 8k buffer.

marmarek:
And generally (_if_ that would be the case, which isn't) no underruns takes priority over low latency.

pulse/pacat-simple-vchan.c
```diff
@@ -210,6 +215,16 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
...
     }
 
+    if (verbose > 1) {
+        pa_usec_t latency;
+        int negative;
+
+        if (pa_stream_get_latency(s, &latency, &negative))
+            pacat_log("pa_stream_get_latency() failed");
+
+        pacat_log("process_playback_data(): vchan data %d max_length %d latency %llu", space_in_vchan, max_length, latency);
```
DemiMarie, line 225:
```suggestion
        if (pa_stream_get_latency(s, &latency, &negative))
            pacat_log("pa_stream_get_latency() failed");
        else
            pacat_log("process_playback_data(): vchan data %d max_length %d latency %llu", space_in_vchan, max_length, latency);
```

otherwise we might log uninitialized stack rubble

DemiMarie (changes requested):
Buffer attribute changes look fine, but if `pa_stream_get_latency()` fails the code will log either wrong data or uninitialized stack rubble.

pulse/pacat-simple-vchan.c
marmarek: Yes, but a) I still want other values logged, and b) thanks to `PA_STREAM_INTERPOLATE_TIMING` that happens basically only once at the very start.

pulse/pacat-simple-vchan.c
```diff
@@ -210,6 +215,16 @@ static void process_playback_data(struct userdata *u, pa_stream *s, size_t max_l
         return;
     }
 
+    if (verbose > 1) {
+        pa_usec_t latency = 0;
+        int negative;
```
DemiMarie, line 220:
```suggestion
        int negative = 0;
```

avoid logging junk from the stack

marmarek: I don't use the value at all, so it doesn't matter.

marmarek:
> system_tests_basic_vm_qrexec_gui@hw1
> 
>     * TC_20_AudioVM_Pulse_debian-12-xfce: [test_223_audio_play_hvm](https://openqa.qubes-os.org/tests/92414#step/TC_20_AudioVM_Pulse_debian-12-xfce/4) (failure)
>       `AssertionError: frequency 977.2385421984841 not in specified range,...`
> 
>     * TC_20_AudioVM_Pulse_fedora-38-xfce: [test_223_audio_play_hvm](https://openqa.qubes-os.org/tests/92414#step/TC_20_AudioVM_Pulse_fedora-38-xfce/4) (failure)
>       `AssertionError: frequency 986.4409885532086 not in specified range,...`
> 
>     * TC_20_AudioVM_Pulse_whonix-workstation-17: [test_220_audio_play_pulseaudio](https://openqa.qubes-os.org/tests/92414#step/TC_20_AudioVM_Pulse_whonix-workstation-17/1) (failure)
>       `AssertionError: frequency 945.4942192569575 not in specified range,...`
> 
>     * TC_20_AudioVM_Pulse_whonix-workstation-17: [test_223_audio_play_hvm](https://openqa.qubes-os.org/tests/92414#step/TC_20_AudioVM_Pulse_whonix-workstation-17/4) (failure)
>       `AssertionError: too short audio, expected 10s, got 8.94485260770975...`

This is a regression. Those tests are about emulated sound card for HVM, so PipeWire agent is not involved. But PulseAudio agent running in a stubdomain is. Depending on a test, both playback and recording failed. I'll inspect logs more closely and experiment with some parameters, but I suspect changing maxlength for the recording stream is a mistake. But it could be also related to migration to PipeWire in dom0, not this PR.

marmarek:
The HVM playback was problem in a test (yet another difference between Pulseaudio and PipeWire). But HVM recording seems to be legit issue.

DemiMarie:
Want me to make the changes conditional on buffer sizes?  The PipeWire agent uses much bigger buffers.

marmarek:
If that will be necessary, I can also add an option for that (qvm-start-daemon knows very well if given pacat process is for the domain itself or its stubdomain).

marmarek:
I have an alternative idea - instead of changing buffer parameters, change how much data is read from the vchan, using this:
```
/** Return the number of bytes requested by the server that have not yet
 * been written.
 *
 * It is possible to write more than this amount, up to the stream's
 * buffer_attr.maxlength bytes. This is usually not desirable, though, as
 * it would increase stream latency to be higher than requested
 * (buffer_attr.tlength).
 *
 * (size_t) -1 is returned on error.
 */
size_t pa_stream_writable_size(const pa_stream *p);
```

DemiMarie:
@marmarek The code already does this for reads initiated by vchan activity.

marmarek:
Hmm, I'm trying to get what is the size (nbytes) parameter to the write callback. In our code it's called `max_length`, but it isn't clear if it's really max size or requested size.

marmarek:
But in that case, my change tlength->maxlength shouldn't matter for playback. But as the failing recording test shows, it does matter for recording, in negative way. I'll rollback the change, but maybe make tlength configurable?

Oh, and also, fragsize 4096 is incompatible with PA's agent 2048 vchan size, and it works only due to blocking vchan calls. I'll try changing to 2048 and see if that helps.

marmarek:
or better change PA's agent vchan to 4096 ?

marmarek:
In any case, this shows rather clearly that prebuf/tlength handling is broken in Pulseaudio...

marmarek:
> The HVM playback was problem in a test (yet another difference between Pulseaudio and PipeWire). But HVM recording seems to be legit issue.

Or maybe another issue with tests? I found this in the log:
```
1708756176.132966 Buffer metrics: maxlength=4194304, tlength=4096, prebuf=4294967295, minreq=4294967295
1708756176.133012 Using sample spec 's16le 2ch 44100Hz', channel map 'front-left,front-right'.
1708756176.150634 Connected to device alsa_output.pci-0000_00_03.0.analog-stereo.monitor (48, not suspended).
1708756176.150793 Recording requested but not allowed
1708756176.180393 Stream started.[K
1708756176.184494 Stream underrun.[K
1708756176.210018 Stream started.[K
1708756177.311861 Stream underrun.[K
1708756177.320785 Stream started.[K
1708756179.361800 Stream underrun.[K
1708756179.362170 Stream started.[K
1708756179.525348 Stream underrun.[K
1708756179.548189 Stream started.[K
1708756180.041450 Stream underrun.[K
1708756180.059797 Stream started.[K
1708756184.109916 Stream underrun.[K
1708756184.129262 Stream started.[K
2024-02-24 01:29:45,889 icon-receiver: connected: test-inst-vm1
1708756186.376039 Setting audio-input to enabled
1708756186.376245 Recording start
app: Connection to qubesd terminated, reconnecting in 1.0 seconds
```

So, recording was enabled quite late in the test, no wonder it recorded shorter sound than intended (or nothing at all). But also, there was a few underruns, might be relevant.
Anyway, this needs a closer look, the issue wasn't there before migrating to PipeWire.

marmarek:
The HVM recording on debian-12 seems to be a pipewire bug (see https://github.com/QubesOS/qubes-issues/issues/8590). It's fixed in newer pipewire (and so the test passes on fedora template). It was "working" before because the test was broken (and I got it fixed in https://github.com/QubesOS/qubes-core-admin/pull/581). I'll exclude it for now.

Remaining thing to check is recording on whonix-workstation template - which runs pulseaudio (not pipewire-pulse).

marmarek:
> Remaining thing to check is recording on whonix-workstation template - which runs pulseaudio (not pipewire-pulse).

I'm struggling with this case. It looks like with pipewire in dom0 it fails, but with pulseaudio in dom0 it works. More specifically, the failing setup is:
- qube based on whonix-workstation-17
- pulseaudio inside the qube (as whonix-ws has), with qubes module unloaded
- virt_mode hvm
- qvm-features audio-model ich6
- running on nested virt (inside KVM)
- dom0 configured to record "monitor" stream by default, and recording allowed for the test qube
- playing in dom0, and recording in VM

I've tried changing various parameters, like fragsize (both down and up), and also updated PA agent inside stubdomain (so it uses 8KiB vchan buffer instead of 2KiB). Nothing helped. What does help is switching dom0 audio daemon back to pulseaudio. The test passes also with (new, fedora-38) pipewire running inside the qube. 

The result is some missing samples, mostly in the middle. And the test fails with: ` AssertionError: too short audio, expected 10s, got 8.36628117913832...` Recorded audio can be fetched from https://openqa.qubes-os.org/tests/92456/file/system_tests-audio-sample-qubes.tests.integ.vm_qrexec_gui.tar.gz (play with `paplay --raw --format=float32le --rate=44100 --channels=1`).

The specific setup above is unusual enough (especially the nested virt) that it shouldn't matter in theory. But very obvious issues in this setup usually correlated with more subtle issues on more usual configurations. And it's usually easier to debug more obvious issue. What I worry here, is that switching audio daemon in dom0 will regress audio recording for non-Linux HVMs (or even Linux, but without qubes packages).

Any ideas what else I can try?

DemiMarie:
> > Remaining thing to check is recording on whonix-workstation template - which runs pulseaudio (not pipewire-pulse).
> 
> I'm struggling with this case. It looks like with pipewire in dom0 it fails, but with pulseaudio in dom0 it works. More specifically, the failing setup is:
> 
> * qube based on whonix-workstation-17
> * pulseaudio inside the qube (as whonix-ws has), with qubes module unloaded
> * virt_mode hvm
> * qvm-features audio-model ich6
> * running on nested virt (inside KVM)
> * dom0 configured to record "monitor" stream by default, and recording allowed for the test qube
> * playing in dom0, and recording in VM
> 
> I've tried changing various parameters, like fragsize (both down and up), and also updated PA agent inside stubdomain (so it uses 8KiB vchan buffer instead of 2KiB). Nothing helped. What does help is switching dom0 audio daemon back to pulseaudio. The test passes also with (new, fedora-38) pipewire running inside the qube.
> 
> The result is some missing samples, mostly in the middle. And the test fails with: ` AssertionError: too short audio, expected 10s, got 8.36628117913832...` Recorded audio can be fetched from https://openqa.qubes-os.org/tests/92456/file/system_tests-audio-sample-qubes.tests.integ.vm_qrexec_gui.tar.gz (play with `paplay --raw --format=float32le --rate=44100 --channels=1`).
> 
> The specific setup above is unusual enough (especially the nested virt) that it shouldn't matter in theory. But very obvious issues in this setup usually correlated with more subtle issues on more usual configurations. And it's usually easier to debug more obvious issue. What I worry here, is that switching audio daemon in dom0 will regress audio recording for non-Linux HVMs (or even Linux, but without qubes packages).
> 
> Any ideas what else I can try?

1. @adrelanos is switching to PipeWire an option?
2. What about using PipeWire in the stubdomain, or (better yet) having QEMU talk to pacat directly, with no audio server involved?

marmarek:
> 1. is switching to PipeWire an option?

That is not the point... I fear what we see here is a symptom of broken audio in HVM in general, that will manifest in other cases, maybe even not Linux.

> 2\. What about using PipeWire in the stubdomain, or (better yet) having QEMU talk to pacat directly, with no audio server involved?

The former I don't like, due to significantly higher memory usage for PipeWire, which is even more critical in stubdomain. The latter might be an option in theory, but given how "well" it went with pipewire agent implementation, I'm skeptical we will improve the situation this way.

marmarek:
Oh, and if I retry the test few more times, it fails with new pipewire in the VM too. But after switching to pulseaudio in dom0 it works every time.

adrelanos:
> 1. @adrelanos is switching to PipeWire an option?

As far as Qubes-Whonix is concerned, yes, very much so. This is unspecific to Qubes-Whonix.

(Non-Qubes-Whonix is already on PipeWire.)

[qubes-whonix `debian/control`](https://github.com/Whonix/qubes-whonix/blob/master/debian/control) contains `Depends:` on  `pulseaudio-qubes`. Happy to change to something else. Please let me know.

DemiMarie:
> Oh, and if I retry the test few more times, it fails with new pipewire in the VM too. But after switching to pulseaudio in dom0 it works every time.

So I think the test matrix is:

1. direction: recording vs playback
2. dom0 audio server: PipeWire + pipewire-pulse vs PulseAudio
3. system: nested virt vs bare hardware
4. guest audio: emulated vs Qubes PV
5. guest audio server: PulseAudio vs PipeWire

Which of these 32 possibilities are good and which are bad?  Is there an openQA run where I can look at the results of all of these?

My understanding is:

1. With PipeWire in both dom0 and domU, QubesOS/qubes-gui-agent-linux#207 is needed to fix a single underrun immediately after the stream is uncorked.  Otherwise, everything works fine on bare hardware, but I have no idea about nested virt.
2. With PulseAudio in dom0, a PulseAudio server bug causes very large latency and problems with the PipeWire agent.  This problem is independent of PipeWire vs PulseAudio in the guest, but the PulseAudio agent doesn’t trigger it, presumably because of extra buffering there.
3. Earlier revisions in this PR tried and failed work around the PulseAudio bug, so the plan is to switch to PipeWire in dom0.
4. Switching to PipeWire in dom0 fixes playback in all cases, but _recording_ is now broken in a subset of cases.

Is this accurate?  I’m somewhat confused as to what is happening here, which is why I am asking for the results in all 32 possible cases.

marmarek:
> Which of these 32 possibilities are good and which are bad? Is there an openQA run where I can look at the results of all of these?

I think it should be possible to collect all the combinations.

Take a most recent run: https://openqa.qubes-os.org/tests/overview?distri=qubesos&version=4.2&build=2024022504-4.2&flavor=pull-requests

click on vm_qrexec_gui tests

> direction: recording vs playback

"rec" or "play" in test name

> dom0 audio server: PipeWire + pipewire-pulse vs PulseAudio

recent runs use pipewire in dom0; for pulseaudio in dom0 you need an earlier run, for example this: https://openqa.qubes-os.org/tests/overview?distri=qubesos&version=4.2&build=2024021900-4.2&groupid=11; but note the audio_rec_unmuted_hvm test wasn't fixed at that time, so it's skipped in many cases)

> system: nested virt vs bare hardware

the one with "@hw1" is on bare hardware, others are nested virt

> guest audio: emulated vs Qubes PV

"hvm" in test name is emulated, others are qubes pv

> guest audio server: PulseAudio vs PipeWire

template:
 - fedora - pipewire >= 1.0
 - debian - pipewire 0.3.something
 - whonix-workstation - pulseaudio
 - whonix-gateway - nothing (and test skipped)

Especially for nested virt, check few jobs (the whole vm_qrexec_gui suite is ran several times with different storage, so you have quite a few data points) - a single passing instance doesn't necessarily mean it's good. Failures are collected in qubesos-bot message in relevant PR (this included), you can check edits history for older reports - that's usually quicker way than manually clicking on all the jobs.

Oh, and one more thing, failure "only silence detected" is not interesting - it's either buggy test, or the known issue in the old pipewire.



> My understanding is:
> 
>     1. With PipeWire in both dom0 and domU, https://github.com/QubesOS/qubes-gui-agent-linux/pull/207 is needed to fix a single underrun immediately after the stream is uncorked. Otherwise, everything works fine on bare hardware, but I have no idea about nested virt.

For PV audio - yes, both bare hw and nested virt is good. And the _single_ underrun at the start is not a problem (looks like an artifact prebuf kicking in), so that PR isn't needed IMO.

> 2\. With PulseAudio in dom0, a PulseAudio server bug causes very large latency and problems with the PipeWire agent.  This problem is independent of PipeWire vs PulseAudio in the guest, but the PulseAudio agent doesn’t trigger it, presumably because of extra buffering there.

Large latencies _and_ a lot of underruns for several seconds after starting playback on nested virt (and some less on bare hw, but still problematic). Neither of those problem affect dom0 PA + domU PA setup, only dom0 PA + domU PW is affected.

> 3\. Earlier revisions in this PR tried and failed work around the PulseAudio bug, so the plan is to switch to PipeWire in dom0.

Yes.

> 4\. Switching to PipeWire in dom0 fixes playback in all cases, but _recording_ is now broken in a subset of cases.

Recording using emulated audio is affected. And on the very last run, I see a single failure of emulated playback (with pulseaudio in domU).

DemiMarie:
> > 1. With PipeWire in both dom0 and domU, https://github.com/QubesOS/qubes-gui-agent-linux/pull/207 is needed to fix a single underrun immediately after the stream is uncorked. Otherwise, everything works fine on bare hardware, but I have no idea about nested virt.
> 
> For PV audio - yes, both bare hw and nested virt is good. And the _single_ underrun at the start is not a problem (looks like an artifact prebuf kicking in), so that PR isn't needed IMO.

I’d still like to eliminate the underrun, mostly because it can hide _real_ underruns (not due to prebuf) which indicate actual problems.  I vaguely remember this being a problem when developing the PipeWire agent.

marmarek:
What I don't like about that PR is second-guessing what dom0 part is doing, including some variable parameters (like buffer size to fill in). That will lead to issues sooner or latter (for example, if we'd reduce pacat buffer sizes at some point in the future, the prefill PR will cause overruns at start). If pacat needs to delay uncork until at least some amount of data is collected, it's the job of pacat to do that, not domU agent. And the prebuf feature does exactly that. I'm not sure why I haven't managed to get it working with real Pulseaudio (it could be a bug in PA, but very well it could be also some other bug in pacat), but with PipeWire in dom0 it seems to work correctly.

PR #141, opened 2024-03-26
pacat-simple-vchan: reconnect on vchan eof

fepitre:
https://github.com/QubesOS/qubes-issues/issues/8975

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -1032,10 +1032,8 @@ int main(int argc, char *argv[])
     struct userdata u;
     pa_glib_mainloop* m = NULL;
     pa_time_event *time_event = NULL;
-    int domid = -1;
     char *pidfile_path;
     int pidfile_fd;
-    int play_watch_fd, rec_watch_fd;
     int i;
 
     memset(&u, 0, sizeof(u));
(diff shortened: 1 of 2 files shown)


pulse/pacat-simple-vchan.c
```diff
@@ -1217,5 +1225,9 @@ int main(int argc, char *argv[])
 
     unlink(pidfile_path);
     close(pidfile_fd);
+
+    if (!u.ret)
+        goto main;
```
marmarek, line 1230: This should be before removing the pidfile, otherwise you end up without it after reconnect.

pulse/pacat-simple-vchan.c
```diff
@@ -1068,27 +1066,30 @@ int main(int argc, char *argv[])
...
     u.ret = 1;
 
     g_mutex_init(&u.prop_mutex);
 
-    u.name = domname;
-
-    u.play_ctrl = libvchan_client_init_async(domid, QUBES_PA_SINK_VCHAN_PORT, &play_watch_fd);
+    u.play_ctrl = libvchan_client_init_async(u.domid, QUBES_PA_SINK_VCHAN_PORT, &u.play_watch_fd);
     if (!u.play_ctrl) {
         perror("libvchan_client_init_async");
```
marmarek, line 1089: This case is now not only initial connect (which, if fails, should result in non-zero exit code), but also in case of reconnect (which may be a clean domain shutdown). In the latter case, exit should be clean. Maybe use `exit(u.ret)` here, which should be 0 in the reconnect case? But also, going for exit directly here will leave pidfile not cleaned up... So, either clean it up here (and similarly in exit() calls below), or add a goto label to the pidfile cleanup at the end of main().

marmarek (changes requested):

pulse/pacat-simple-vchan.c
fepitre: I've made a proposal by simply calling a function instead of adding another goto.

pulse/pacat-simple-vchan.c
```diff
@@ -1159,6 +1166,8 @@ int main(int argc, char *argv[])
     /* Run the main loop */
     g_main_loop_run (u.loop);
 
+    cleanup_pidfile(pidfile_path, pidfile_fd);
```
marmarek, line 1169: Here you remove the pidfile on reconnect. You have a call at the very end of the main function already, remove this one.

marmarek (changes requested):

marmarek (approved):

PR #144, opened 2024-06-23
Add window_background_color config setting

Euwiiwueir:
Note: in QubesOS/qubes-issues#9304 I wrote:

> My instinct is that an option of black or white is good enough; allowing the color to be an arbitrary hue perhaps grants this issue more attention than it deserves. Also, the implementation is simpler. But I defer to others.

I changed my mind about this in the meantime, I like dark blue, and in this change I allow the setting to be an arbitrary user-selected color. This obligates some error checking around `strtoul` and `XAllocNamedColor`. (An input sanitizing step upstream in memory-safe `qvm-features` -- QubesOS/qubes-issues#7730 -- would be a good complement.) If the user-selected color is invalid, `qubes-guid` bails with an error logged to `/var/log/qubes/guid.<vm>.log`.

I think this is ok? But let me know.

Fixes QubesOS/qubes-issues#9304

### DIFF
--- a/gui-daemon/guid.conf
+++ b/gui-daemon/guid.conf
@@ -70,6 +70,13 @@ global: {
   #
   # trayicon_mode = "border1";
 
+  # Set the fill color to be shown in a window when content is pending or
+  # unavailable. This is rarely visible except very briefly. Possible values are
+  # a color name (see: /etc/X11/rgb.txt) or a specification in format 0xRRGGBB.
+  # When running a dark-styled desktop theme, "black" is recommended.
+  #
+  # window_background_color = "white";
+
   # Timeout when waiting for qubes-gui-agent
   #
   # startup_timeout = 45;
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -283,27 +283,59 @@ static int x11_io_error_handler(Display * dpy)
     exit(1);
 }
 
+/*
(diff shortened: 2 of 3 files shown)


Euwiiwueir:
> Suggestion: If you intend for this PR to fix/close issue [QubesOS/qubes-issues#9304](https://github.com/QubesOS/qubes-issues/issues/9304) and would like it to be linked to that issue automatically, use:
> 
> ```
> Fixes QubesOS/qubes-issues#9304
> ```

Thanks. My hesitation was that there are two pull requests that only jointly resolve the issue. Should one or both have `Fixes`? Well, I suppose it doesn't matter too much either way.

marmarek:
On Mon, Jun 24, 2024 at 08:52:11AM -0700, Euwiiwueir wrote:
> Thanks. My hesitation was that there are two pull requests that only jointly resolve the issue. Should one or both have `Fixes`? Well, I suppose it doesn't matter too much either way.

Better both. This way, both resulting packages will be linked to the
issue.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

marmarek (approved):

fepitre:
I think an update should be done for documentation: https://www.qubes-os.org/doc/config-files/

Euwiiwueir:
Here is the issue: https://github.com/QubesOS/qubes-issues/issues/9306

PR #147, opened 2024-08-20
device identity

piotrbartman:
implements QubesOS/qubes-issues/issues/9325


marmarek (approved):
I think attach/detach in domain-start/domain-shutdown events can be called via await directly instead of ensure_future, but the current version is okay too, so lets merge it as is for now

PR #148, opened 2024-09-25
Fix: make shmoverride.so initfirst

CertainLach:
After I have rebased my qubesos nixos dom0 on latest nixos version, I have encountered Xwayland/Xorg failing with segfault.

The cause of segfault was fstat hook being called before library constructor:
image

image

image

This was not caused by constructor not finding fstat with dlsym, I have checked that no logs being emitted by this library (and double-checked by using bpftrace to catch all fprints to stderr), constructor defenitely was not being called.

I'm not sure why, but `-z initfirst` linker flag fixes this constructor call ordering issue, making constructor being run before fsync, thus fixing the segfault, and restoring the Xwayland/Xorg operation.

I can provide more info/logs about this issue, but my knowledge about dynamic linking is limited, and I have no idea what might be useful here.

### DIFF
--- a/shmoverride/shmoverride.c
+++ b/shmoverride/shmoverride.c
@@ -72,6 +72,8 @@ static int (*real_munmap) (void *shmaddr, size_t len);
 static int (*real_fstat64) (VER_ARG int fd, struct stat64 *buf);
 static int (*real_fstat)(VER_ARG int fd, struct stat *buf);
 
+static int try_init(void);
+
 static struct stat global_buf;
 static int gntdev_fd = -1;
 
@@ -84,7 +86,7 @@ static int xc_hnd;
 static xengnttab_handle *xgt;
 static char __shmid_filename[SHMID_FILENAME_LEN];
 static char *shmid_filename = NULL;
-static int idfd = -1, display = -1;
+static int idfd = -1, display = -1, init_called = 0;
 
 static uint8_t *mmap_mfns(struct shm_args_hdr *shm_args) {
     uint8_t *map;
@@ -150,6 +152,8 @@ ASM_DEF(void *, mmap,
         real_fstat64 = FSTAT64;
         real_fstat = FSTAT;
     }
+
+    try_init();
 
 #if defined MAP_ANON && defined MAP_ANONYMOUS && (MAP_ANONYMOUS) != (MAP_ANON)
 # error header bug (def mismatch)
@@ -217,6 +221,9 @@ ASM_DEF(int, munmap, void *addr, size_t len)
 {
     if (len > SIZE_MAX - XC_PAGE_SIZE)
         abort();
+
+    try_init();
+
     const uintptr_t addr_int = (uintptr_t)addr;
     const uintptr_t rounded_addr = addr_int & ~(uintptr_t)(XC_PAGE_SIZE - 1);
     return real_munmap((void *)rounded_addr, len + (addr_int - rounded_addr));
@@ -438,6 +445,7 @@ static int assign_off(off_t *off) {
 
 #define STAT(id)                                          \
 ASM_DEF(int, f ## id, int filedes, struct id *buf) {      \
+    try_init();                                           \
     int res = real_f ## id(VER filedes, buf);             \
     if (res ||                                            \
         !S_ISCHR(buf->st_mode) ||                         \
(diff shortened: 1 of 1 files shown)


marmarek:
I guess it's an interaction with a constructor from another library. Does NixOS use LD_PRELOAD (with something that has a constructor) and/or patches some of the library used by Xwayland to add/modify a constructor?

CertainLach:
Nope, the fsync call iscoming from llvm, which is used by mesa. Maybe llvm/mesa update caused this?

CertainLach:
As I can see, old llvm version was initialized before shmoverride too, but it wasn't calling fstat before.
Now it is calling fstat, and because it is initializing before shmoverride - everything crashes.

I believe it is correct to have -z initfirst linker flag.

marmarek:
> system_tests_gui_tools
> 
>     * startup: [unnamed test](https://openqa.qubes-os.org/tests/113326#step/startup/11) (unknown)
> 
>     * startup: [Failed](https://openqa.qubes-os.org/tests/113326#step/startup/12) (test died)
>       `# Test died: no candidate needle with tag(s) 'nm-connection-establi...`

> (and all the other failures like this)

This looks to be caused by this PR. I see the following in the log:
```
Missing /var/run/qubes/shm.id.0; run X with preloaded shmoverride
```

CertainLach:
That's strange, I have tested this change on QubesOS with no problems, it should not affect library loading at all, it only affects initialization.

marmarek:
Well, maybe it is loaded, but constructor is not called? Or it fails? I don't see any errors from it in Xorg.0.log. You can find more logs in "logs & assets" tab on openqa (click any of those failed tests in the comment)

CertainLach:
Yep, also a possibility, I'll research.

CertainLach:
For some reason, it is not working when built using QubesOS build system, and I have no idea why.
For now, I have implemented safer version of this patch with bigger performance penality, which works everywhere.

I'll include it with my NixOS patchset until I figure out what would be best way to workaround this issue.

In NixOS, I think it would be easier to directly patch Xorg, instead of hooking its functions.
Or use hotpatch instead of loader magic, as it will also work here too.

CertainLach:
Huh, both current and previous version of this patch works fine on NixOS with Xwayland

CertainLach:
> did unsetenv("LD_PRELOAD") not work?

Maybe it did, but some other initializer spawned processes before try_init was called?
Have you kept the constructor function? Because it is being called on NixOS, earlier try_init calls are only inserted in the case if some other library initializer (e.g llvm) wants to use fstat/mmap

marmarek:
> Maybe it did, but some other initializer spawned processes before try_init was called?

Looking at what got spawned, it's unlikely.

> Have you kept the constructor function?

Yes.

CertainLach:
Have you used only one patch (try_init), or both at the same time (try_init + initfirst)?

try_init doesn't require initfirst to be set

CertainLach:
How about using patchelf to reroute calls to fstat/mmap from wanted library to shmoverride.so directly? (Just --add-needed shmoverride.so, --rename-dynamic-symbols fstat -> shmoverride_fstat, etc)
I'm not sure how well that may work in rpm-based distributions, but this is the route I would like to take in NixOS

marmarek:
Could it be that `unsetenv` with initfirst happens too early and env gets re-loaded by some other constructor (maybe in glibc)?
I guess it's harmless as long as X server doesn't spawn further processes (based on my log, it's just xkbcomp, so maybe it doesn't happen if keyboard layout is not changed?).

marmarek:
I'm testing things in #151

marmarek:
> Could it be that `unsetenv` with initfirst happens too early and env gets re-loaded by some other constructor (maybe in glibc)?

I think this is the case, the `getenv` doesn't work with `-z initfirst` either.

marmarek:
And #152 has extended config, but due to getenv not working with initfirst, it's broken.

marmarek:
So, all in all, IMO the slower (current) version is better. The overhead shouldn't be that big. In fact, it looks like the compiler is smart enough to inline the `if (init_called)` part.

CertainLach:
> In fact, it looks like the compiler is smart enough to inline the if (init_called) part.

I was checking on godbolt, and __builtin_expect makes this optimization always work, without it - I haven't seen inlining here.

marmarek:
Is there anything left to change here? If not, can you make it non-draft?

marmarek (approved):

PR #149, opened 2024-10-01
Add event buffering for cloaking user input patterns

ArrayBolt3:
# Goal

Implement the functionality of [kloak](https://github.com/Whonix/kloak) (a tool designed to hide biometric behavior patterns in keystrokes and mouse movements) in qubes-gui-daemon. This PR will implement the functionality requested in https://github.com/QubesOS/qubes-issues/issues/1850 and fleshed out further in https://github.com/QubesOS/qubes-issues/issues/8541. It will also close https://github.com/QubesOS/qubes-issues/issues/8534 as it will no longer be necessary.

# TODOs

* Test rigorously on Qubes R4.3 (an earlier iteration of the code has been smoke-tested on Qubes R4.2, this hasn't been tested at all on R4.3 yet)

# Fixed TODOs:

* Figure out why the domU window occasionally freezes until another input event is sent - we aren't buffering info coming from domU to dom0 so why this is happening is a mystery to me, and something for later investigation. (Solved, https://github.com/QubesOS/qubes-gui-daemon/pull/149#issuecomment-2387143732)
* Potentially change how events are treated (do some events have to operate in pairs for best results?). (Lots of X events are now not buffered in the latest implementation. Only ones that look valuable to buffer are buffered.)
* Make the delay duration user-configurable (right now it's hardcoded to 150 milliseconds). (Implemented.)
* Allow configuring event delay duration for individual VMs buffering (right now it is applied equally to all VMs). (Implemented.)
* Get the configuration code working and test it. (Solved, this ended up requiring a change to [core-admin-client](https://forum.qubes-os.org/t/cannot-change-qubes-gui-daemon-settings-using-qvm-features/29345/3) which I will be submitting as a separate PR.)
* Ensure all new code adheres to Qubes OS standards (didn't have time to finish that up) (should be done now)

# Rationale

Kloak, the inspiration for this PR, is a user input buffering and obfuscation tool. It intercepts keyboard and mouse events at the evdev layer, holds them in a queue for release at a later scheduled time, then releases them to the applications they were intended for periodically. By adding random noise into the user's input patterns, kloak aims to make otherwise recognizable patterns in user behavior (such as keystroke rhythm and mouse movement patterns) too erratic to be used as a method of identifying the user. This is potentially very useful especially for Whonix Workstation domUs, as it denies an adversary access to a remarkably effective biometric fingerprinting mechanism they could otherwise access without specialized tools.

Kloak is currently able to operate directly in Qubes domUs if (and only if!) `gui-agent-virtual-input-device` is enabled for the domU in question. Even in these instances, only keyboard events are anonymized, and additionally the domU must have an evdev X driver installed. This is less than ideal from a functionality standpoint, and as @DemiMarie has explained in https://github.com/QubesOS/qubes-issues/issues/8541 it will eventually stop working entirely. There's also the possibility of malware compromise in the domU resulting in the deanonymization of the user. For these reasons, enabling the use of evdev in domUs and running kloak in the domU is not a good solution.

The other obvious option is to run kloak directly in dom0. This has several disadvantages:

* kloak can now potentially wreak havoc on the user's ability to use their computer. If a bug in kloak locks up the keyboard, or the user does something inadvisable like setting a 20-second event delay, regaining control of the system could be difficult or impossible without doing a hard reset (or worse, booting an external USB in order to chroot into dom0 and disable kloak).
* Application of kloak's functionality becomes all-or-nothing - you either anonymize all keyboard and mouse input everywhere, or you anonymize none of it. This could make management of dom0 annoying with larger delay times, and it could prevent the user from making use of applications or websites that require input pattern telemetry to function (such as some bank websites).
* kloak's configuration options similarly apply globally. One might want a comfortable delay of only 25ms in a domU they expect to be safe, but wish to use an extremely long one like 1000ms in a domU they believe is compromised and actively exfiltrating data. With kloak running directly in dom0, this is impossible.

This PR implements a third option - *inserting the functionality of kloak directly into qubes-gui-daemon.* kloak upstream never needs to be involved, only the functionality of it must be. This functionality I have termed "event buffering", and as this implementation works with X server events I have called it "event buffering", or "ebuf" for short (which is the term used for it in the code). Previously I had called this "X event buffering" and used "xbuf" for short, but as @3hhh pointed out that name would become inaccurate when this is ported to Wayland, so I changed it to "ebuf" so as to make the name be display server agnostic.

By working inside the GUI daemon, the following advantages are gained:

* No evdev support needed at all, we can work with X events instead.
* The amount of additional code needed is smaller.
* Per-VM application and configuration of event buffering is now possible - some VMs can use a small, comfortable delay, others can use a very long one, and others can skip delays entirely.
* Even if something goes very wrong and buffering prevents the user from inputting anything into any Qube, the user retains control of dom0 and can recover their system from there.
* A compromised domU with event buffering enabled will be less likely to leak valuable biometric info to the malware within the VM.

# How it works

Most of the code should be fairly self-explanatory. In a nutshell, we use a tail queue to store a list of delayed, scheduled X events. As events come from dom0 to a domU, they are captured, scheduled for release at a later time, and thrown into the queue. Events in the queue are regularly checked to see if their scheduled release time has arrived, and those events are released when appropriate. The scheduler inserts some random noise into the delays, making it difficult to uniquely identify the user's typing and mouse movement/usage patterns.

By default, event buffering is disabled and all events are passed through without buffering. To enable it, one must use `qvm-features` to set `gui-ebuf-max-delay` to a value greater than 0. It is worth noting that 0 is interpreted not as a "don't add any delay when buffering events", but rather it is interpreted as "don't buffer events at all". This configuration feature **does not work** without the `ebuf_max_events` setting being added to the list of GUI daemon configuration settings in qubes-core-admin-client. The pull request for that is at https://github.com/QubesOS/qubes-core-admin-client/pull/309.

This PR needs more testing (especially on Qubes R4.3), but it is solid enough that I feel comfortable asking for a review on it. Thanks for your help!

### DIFF
--- a/gui-common/txrx-vchan.c
+++ b/gui-common/txrx-vchan.c
@@ -38,8 +38,6 @@ int vchan_is_closed = 0;
  */
 int double_buffered = 1;
 
-static int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd);
-
 void vchan_register_at_eof(void (*new_vchan_at_eof)(void)) {
     vchan_at_eof = new_vchan_at_eof;
 }
@@ -99,7 +97,7 @@ int read_data(libvchan_t *vchan, char *buf, int size)
     int ret;
     while (written < size) {
         while (!libvchan_data_ready(vchan))
-            wait_for_vchan_or_argfd_once(vchan, -1);
+            wait_for_vchan_or_argfd_once(vchan, -1, 1000);
         ret = libvchan_read(vchan, buf + written, size - written);
         if (ret <= 0)
             handle_vchan_error(vchan, "read data");
@@ -109,7 +107,7 @@ int read_data(libvchan_t *vchan, char *buf, int size)
     return size;
 }
 
-static int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd)
+int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd, int timeout)
 {
     int ret;
     write_data(vchan, NULL, 0);    // trigger write of queued data, if any present
@@ -117,7 +115,7 @@ static int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd)
         { .fd = libvchan_fd_for_select(vchan), .events = POLLIN, .revents = 0 },
         { .fd = fd, .events = POLLIN, .revents = 0 },
     };
-    ret = poll(fds, fd == -1 ? 1 : 2, 1000);
+    ret = poll(fds, fd == -1 ? 1 : 2, timeout);
     if (ret < 0) {
         if (errno == EINTR)
             return -1;
@@ -138,12 +136,5 @@ static int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd)
         // clear libvchan_fd pending state 
         libvchan_wait(vchan);
     }
-    return ret;
-}
-
-int wait_for_vchan_or_argfd(libvchan_t *vchan, int fd)
-{
-    int ret;
-    while ((ret=wait_for_vchan_or_argfd_once(vchan, fd)) == 0);
     return ret;
 }
--- a/gui-daemon/guid.conf
+++ b/gui-daemon/guid.conf
@@ -85,4 +85,10 @@ global: {
   # 256 to 256000 characters. Default is 64000 characters.
   #
   # max_clipboard_size = 64000
+
+  # Maximum delay in milliseconds for event buffering (used for obfuscating
+  # biometric data that could be obtained by observing keyboard and mouse
+  # patterns). Set to 0 to disable event buffering entirely.
+  #
+  # events_max_delay = 0;
 }
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -4477,7 +4477,7 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
             g->disable_override_redirect = 0;
         else {
             fprintf(stderr,
-                    "unsupported value ‘%s’ for override_redirect (must be ‘disabled’ or ‘allow’)\n",
+                    "unsupported value '%s' for override_redirect (must be 'disabled' or 'allow')\n",
                     value);
             exit(1);
         }
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -37,6 +37,8 @@
 #include <sys/wait.h>
 #include <sys/resource.h>
 #include <sys/uio.h>
+#include <sys/queue.h>
+#include <sys/random.h>
 #include <signal.h>
 #include <poll.h>
 #include <errno.h>
@@ -2592,11 +2594,85 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
 
 }
 
-/* dispatch local Xserver event */
-static void process_xevent(Ghandles * g)
+/* get current time */
+static int64_t ebuf_current_time_ms()
 {
-    XEvent event_buffer;
-    XNextEvent(g->display, &event_buffer);
+    int64_t timeval;
+    struct timespec spec;
+    clock_gettime(CLOCK_MONOTONIC, &spec);
+    timeval = (((int64_t)spec.tv_sec) * 1000LL) + (((int64_t)spec.tv_nsec) / 1000000LL);
+    return timeval;
+}
+
+/* get random delay value */
+static uint32_t ebuf_random_delay(uint32_t upper_bound, uint32_t lower_bound)
+{
+    uint32_t maxval;
+    uint32_t randval;
+    size_t randsize;
+    union ebuf_rand ebuf_rand_data;
+
+    if (lower_bound >= upper_bound) {
+        if (lower_bound > upper_bound) {
+            fprintf(stderr,
+                    "Bug detected - lower_bound > upper_bound, events may get briefly stuck");
+        }
+        return upper_bound;
+    }
+
+    maxval = upper_bound - lower_bound + 1;
+
+    do {
+        randsize = getrandom(ebuf_rand_data.raw, sizeof(uint32_t), 0);
+        if (randsize != sizeof(uint32_t))
+            continue;
+    } while (ebuf_rand_data.val > UINT32_MAX - ((uint32_t)((((uint64_t)UINT32_MAX) + 1) % maxval)));
+
+    randval = ebuf_rand_data.val % maxval;
+    return lower_bound + randval;
+}
+
+/* queue input event */
+static void ebuf_queue_xevent(Ghandles * g, XEvent xev)
+{
+    int64_t current_time;
+    uint32_t random_delay;
+    struct ebuf_entry *new_ebuf_entry;
+    uint32_t lower_bound;
+
+    current_time = ebuf_current_time_ms();
+
+    /* 
+     * Each event is scheduled by taking the current time and adding a delay.
+     * We do not want events later in the queue having a release timestamp
+     * that is *less* than an event earlier in the queue. This means that
+     * whatever delay we add *must* be at least enough to give a release
+     * timestamp larger than the one generated last time. To facilitate this,
+     * we set lower_bound to the scheduled release time of the last event
+     * minus the current time. Some sanity checks are included to make sure
+     * lower_bound is never less than 0 or greater than ebuf_max_delay.
+     */
+    lower_bound = min(max(g->ebuf_prev_release_time - current_time, 0), g->ebuf_max_delay);
+
+    random_delay = ebuf_random_delay(g->ebuf_max_delay, lower_bound);
+    new_ebuf_entry = malloc(sizeof(struct ebuf_entry));
+    if (new_ebuf_entry == NULL) {
+        perror("Could not allocate ebuf_entry:");
+        exit(1);
+    }
+    if (current_time > 0 && random_delay > (INT64_MAX - current_time)) {
+        fprintf(stderr, "Event scheduler overflow detected, cannot continue");
+        exit(1);
+    }
+    new_ebuf_entry->time = current_time + random_delay;
+    new_ebuf_entry->xev = xev;
+    TAILQ_INSERT_TAIL(&(g->ebuf_head), new_ebuf_entry, entries);
+    g->ebuf_prev_release_time = new_ebuf_entry->time;
+}
+
+/* dispatch local Xserver event */
+static void process_xevent_core(Ghandles * g, XEvent event_buffer)
+{
     switch (event_buffer.type) {
     case KeyPress:
     case KeyRelease:
@@ -2652,6 +2728,40 @@ static void process_xevent(Ghandles * g)
         }
         break;
     default:;
+    }
+}
+
+/* dispatch queued events */
+static void ebuf_release_xevents(Ghandles * g)
+{
+    int64_t current_time;
+    struct ebuf_entry *current_ebuf_entry;
+
+    current_time = ebuf_current_time_ms();
+    while ((current_ebuf_entry = TAILQ_FIRST(&(g->ebuf_head)))
+        && (current_time >= current_ebuf_entry->time)) {
+        XEvent event_buffer = current_ebuf_entry->xev;
+        process_xevent_core(g, event_buffer);
+        TAILQ_REMOVE(&(g->ebuf_head), current_ebuf_entry, entries);
+        free(current_ebuf_entry);
+    }
+    current_ebuf_entry = TAILQ_FIRST(&(g->ebuf_head));
+    if (current_ebuf_entry == NULL) {
+        g->ebuf_next_timeout = VCHAN_DEFAULT_POLL_DURATION;
+    } else {
+        g->ebuf_next_timeout = (int)(current_ebuf_entry->time - current_time);
+    }
+}
+
+/* handle or queue local Xserver event */
+static void process_xevent(Ghandles * g)
+{
+    XEvent event_buffer;
+    XNextEvent(g->display, &event_buffer);
+    if (g->ebuf_max_delay > 0) {
+        ebuf_queue_xevent(g, event_buffer);
+    } else {
+        process_xevent_core(g, event_buffer);
     }
 }
 
@@ -4482,6 +4592,18 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
             exit(1);
         }
     }
+
+    if ((setting =
+         config_setting_get_member(group, "events_max_delay"))) {
+        int delay_val = config_setting_get_int(setting);
+        if (delay_val < 0 || delay_val > 5000) {
+            fprintf(stderr,
+                    "unsupported value '%d' for events_max_delay (must be >= 0 and <= 5000)",
+                    delay_val);
+            exit(1);
+        }
+        g->ebuf_max_delay = delay_val;
+    }
 }
 
 static void parse_config(Ghandles * g)
@@ -4604,6 +4726,8 @@ int main(int argc, char **argv)
     /* parse cmdline, possibly overriding values from config */
     parse_cmdline(&ghandles, argc, argv);
     get_boot_lock(ghandles.domid);
+    /* init event queue */
+    TAILQ_INIT(&(ghandles.ebuf_head));
 
     if (!ghandles.nofork) {
         // daemonize...
@@ -4799,8 +4923,15 @@ int main(int argc, char **argv)
                 handle_message(&ghandles);
                 busy = 1;
             }
+            if (ghandles.ebuf_max_delay > 0) {
+                ebuf_release_xevents(&ghandles);
+            }
         } while (busy);
-        wait_for_vchan_or_argfd(ghandles.vchan, xfd);
+        if (ghandles.ebuf_max_delay > 0) {
+            wait_for_vchan_or_argfd_once(ghandles.vchan, xfd, ghandles.ebuf_next_timeout);
+        } else {
+            wait_for_vchan_or_argfd_once(ghandles.vchan, xfd, VCHAN_DEFAULT_POLL_DURATION);
+        }
     }
     return 0;
 }
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -4608,7 +4608,7 @@ int main(int argc, char **argv)
     if (!ghandles.nofork) {
         // daemonize...
         if (pipe(pipe_notify) < 0) {
-            perror("canot create pipe:");
+            perror("cannot create pipe:");
             exit(1);
         }
 
--- a/gui-daemon/xside.h
+++ b/gui-daemon/xside.h
@@ -72,6 +72,8 @@
 
 #define MAX_SCREENSAVER_NAMES 10
 
+#define VCHAN_DEFAULT_POLL_DURATION 1000
+
 #ifdef __GNUC__
 #  define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
 #else
@@ -82,6 +84,7 @@
 #include <stdbool.h>
 #include <assert.h>
 #include <unistd.h>
+#include <sys/queue.h>
 #include <libvchan.h>
 #include <X11/Xlib.h>
 #include <xcb/xcb.h>
@@ -137,6 +140,17 @@ struct extra_prop {
     int format; /* data format (8, 16, 32) */
     void *data; /* actual data */
     int nelements; /* data size, in "format" units */
+};
+
+struct ebuf_entry {
+    XEvent xev;
+    int64_t time;
+    TAILQ_ENTRY(ebuf_entry) entries;
+};
+
+union ebuf_rand {
+    uint32_t val;
+    char raw[sizeof(uint32_t)];
 };
 
 /* global variables
@@ -240,6 +254,11 @@ struct _global_handles {
     int xen_fd; /* O_PATH file descriptor to /dev/xen/gntdev */
     int xen_dir_fd; /* file descriptor to /dev/xen */
     bool permit_subwindows : 1; /* Permit subwindows */
+    uint32_t ebuf_max_delay;
+    /* ebuf state */
+    TAILQ_HEAD(tailhead, ebuf_entry) ebuf_head;
+    int64_t ebuf_prev_release_time;
+    int ebuf_next_timeout;
 };
 
 typedef struct _global_handles Ghandles;
--- a/include/txrx.h
+++ b/include/txrx.h
@@ -33,7 +33,7 @@ int read_data(libvchan_t *vchan, char *buf, int size);
     x.untrusted_len = sizeof(y); \
     real_write_message(vchan, (char*)&x, sizeof(x), (char*)&y, sizeof(y)); \
     } while(0)
-int wait_for_vchan_or_argfd(libvchan_t *vchan, int fd);
+int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd, int timeout);
 void vchan_register_at_eof(void (*new_vchan_at_eof)(void));
 
 #endif /* _QUBES_TXRX_H */


DemiMarie:
Please use `getrandom()` instead of ISAAC.

ArrayBolt3:
@DemiMarie I can do that, but that will drain the system's entropy sources at a fairly constant rate (every mouse movement, keystroke, etc. will result in entropy drain). Theoretically that could weaken any automatically generated encryption keys used for things like HTTPS and the like. Using ISAAC requires only a single initial use of system entropy.

If entropy drain isn't a concern, then I'm happy to swap it out.

ArrayBolt3:
ISAAC removed, `getrandom()`-based delay mechanism implemented.

I also tracked down the source of the GUI freeze bug - turned out to be because of `gui-common/txrx-vchan.c:wait_for_vchan_or_argfd`.

```c
int wait_for_vchan_or_argfd(libvchan_t *vchan, int fd) 
{
    int ret;
    while ((ret=wait_for_vchan_or_argfd_once(vchan, fd)) == 0);
    return ret;
}
```

This was apparently busy-waiting for something to happen and thus keeping queued events from ever being released until the user did something like press a key or move the mouse. I used a hack to make `wait_for_vchan_or_argfd` non-blocking, but I'm not totally sure that's going to be acceptable in the long run since this will probably significantly increase the CPU usage of `qubes-guid`. If this is an acceptable solution, `wait_for_vchan_or_argfd` should just be removed and the underlying function `wait_for_vchan_or_argfd_once` should be made public so that `xside.c` can use it directly.

My commit also attempts to implement configuration support - I don't see any reason why the code *shouldn't* work, but I haven't yet gotten it to work on my machine as I seem to be having trouble setting a custom VM feature properly.

DemiMarie:
> @DemiMarie I can do that, but that will drain the system's entropy sources at a fairly constant rate (every mouse movement, keystroke, etc. will result in entropy drain). Theoretically that could weaken any automatically generated encryption keys used for things like HTTPS and the like. Using ISAAC requires only a single initial use of system entropy.
> 
> If entropy drain isn't a concern, then I'm happy to swap it out.

Entropy drain doesn’t actually exist.  The entropy obtained from the `getrandom()` syscall cannot be used to derive the internal state of the Linux CSPRNG.

ArrayBolt3:
I believe this is now ready for review.

ArrayBolt3:
Pull request for configuring X event buffering: https://github.com/QubesOS/qubes-core-admin-client/pull/309

gui-daemon/xside.c
```diff
@@ -2449,6 +2508,55 @@ static void process_xevent(Ghandles * g)
...
+            // Buffered:
+            //     KeyPress
+            //     KeyRelease
+            //     ButtonPress
+            //     ButtonRelease
+            //     MotionNotify
+            //     FocusIn
+            //     FocusOut
+            xbuf_queue_xevent(event_buffer);
+        }
```
marmarek, line 2554: This potentially may result in reordering events. For example sending ButtonPress after LeaveNotify (so, application got info that mouse left its area, and then gets a click). Or ButtonPress delayed after ConfigureNotify, which in some cases may result in clicking wrong window (imagine user clicked in a window, maybe exactly to resize it and then quickly made it smaller, while the click event was still queued). There may be other problematic scenarios...

I don't know what is the best solution. Buffering all events is probably the safest option, but it may be annoying to the user. Or maybe buffer all events only if some (of the selected types) are still queued?

gui-daemon/xside.c
```diff
@@ -4242,6 +4350,11 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
             exit(1);
         }
     }
+
+    if ((setting =
+         config_setting_get_member(group, "xbuf_max_delay"))) {
+        xbuf_max_delay = config_setting_get_int(setting);
```
marmarek, line 4356: This should do a range check. Negative values would underflow, but also too large value will cause issues.

gui-daemon/xside.c
```diff
@@ -4559,8 +4674,11 @@ int main(int argc, char **argv)
                 handle_message(&ghandles);
                 busy = 1;
             }
+            if (xbuf_max_delay > 0) {
+                xbuf_release_xevents(&ghandles);
+            }
         } while (busy);
-        wait_for_vchan_or_argfd(ghandles.vchan, xfd);
+        wait_for_vchan_or_argfd_once(ghandles.vchan, xfd);
```
marmarek, line 4681: As said in the other comment, this can get an extra argument with the time to the next buffered event (if any).

marmarek (changes requested):

gui-daemon/xside.c
```diff
@@ -2386,11 +2395,61 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    return (spec.tv_sec) * 1000 + (spec.tv_nsec) / 1000000;
+}
+
+/* get double-bounded random number */
+static long xbuf_random_between(long lower, long upper)
+{
+    long maxval;
+    long randval;
+    size_t randsize;
+    if (lower >= upper)
```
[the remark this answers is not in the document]
ArrayBolt3: It's more of a sanity check to avoid allowing invalid data to be passed in. The condition there should *never* trigger. This is more-or-less how it was written originally in kloak, but it might be worth making it abort the GUI daemon if it does occur?

gui-daemon/xside.c
```diff
@@ -2386,11 +2395,61 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+
+    maxval = upper - lower + 1;
+    if (maxval > UINT32_MAX)
+        return UINT32_MAX;
+    
+    do {
+        randsize = getrandom(xbuf_rand_data.raw, sizeof(uint64_t), 0);
+        if (randsize != sizeof(uint64_t))
+            continue;
+    } while (xbuf_rand_data.val >= UINT64_MAX - (UINT64_MAX % maxval));
```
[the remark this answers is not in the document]
ArrayBolt3: Probably not needed, I'll rewrite to only use 32 bits.

gui-daemon/xside.c
ArrayBolt3: I had originally written this to buffer all events. I then changed it since it seemed like it might cause issues to do so, but in retrospect X works over SSH which introduces lag all on its own so really it should not be a problem unless there's functionality in the GUI daemon that would render some of these events to be time-sensitive.

gui-common/txrx-vchan.c
```diff
@@ -109,15 +107,15 @@ int read_data(libvchan_t *vchan, char *buf, int size)
...
+int wait_for_vchan_or_argfd_once(libvchan_t *vchan, int fd)
 {
     int ret;
     write_data(vchan, NULL, 0);    // trigger write of queued data, if any present
     struct pollfd fds[] = {
         { .fd = libvchan_fd_for_select(vchan), .events = POLLIN, .revents = 0 },
         { .fd = fd, .events = POLLIN, .revents = 0 },
     };
-    ret = poll(fds, fd == -1 ? 1 : 2, 1000);
+    ret = poll(fds, fd == -1 ? 1 : 2, 1); // short timeout necessary for xbuf
```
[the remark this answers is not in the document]
ArrayBolt3: Implemented as requested.

gui-daemon/xside.c
```diff
@@ -112,6 +114,13 @@ static Ghandles ghandles;
 
 #define ignore_result(x) do { __typeof__(x) __attribute__((unused)) _ignore=(x); } while (0)
 
+/* xbuf state and definitions */
+TAILQ_HEAD(tailhead, xbuf_entry) xbuf_head;
+union xbuf_rand xbuf_rand_data;
+long xbuf_lower_bound = 0;
+long xbuf_prev_release_time = 0;
+long xbuf_max_delay = 0;
```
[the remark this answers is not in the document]
ArrayBolt3: Moved "bad" global variables to local variables, moved remaining global variable to Ghandles.

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Added a warning that will be printed to the log if this condition is hit. Worst case scenario it should only hang up events from reaching the VM for a short period of time so I didn't consider it worth bringing down the entire GUI daemon over, but it would be a bug and so it is now logged if it ever gets triggered.

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Reimplemented to use 32-bit values where appropriate. Some 64-bit values are still in use in the code since 64-bit timestamps are what are being used.

Worthy of note, I do *not* do overflow checking when working with signed timestamps. I don't expect calculations with timestamps to overflow, but it might be worth doing bounds checking anyway since the code does use a monotonic timer and in theory that timer *could* return really large values for some reason.

gui-daemon/xside.c
ArrayBolt3: Changed this to buffer all events. I did *not* check all the event handlers to ensure that they weren't time-sensitive, however in practice using Firefox, vmonaco's device fingerprinting test, and working a bit in the terminal everything *seems* to work so far. Still, this needs further auditing and testing.

gui-daemon/xside.c
ArrayBolt3: Range check added, values less than 0 or greater than 5000 are now forbidden. (5000 is already a delay high enough to make recovery painful and slow, so I don't expect anyone will want to set a delay higher than that.)

gui-daemon/xside.c
ArrayBolt3: Resolved as mentioned in the other comment.

3hhh:
It's nice to see you work on this @ArrayBolt3, much appreciated!

I also believe that this may be relatively easy to port to Wayland.
@DemiMarie: What do you think?

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Double-checked the random number handling code. I don't think any case can overflow (signed or otherwise) except for this line:

    new_xbuf_entry->time = current_time + random_delay;

I'll fix this in the next update by explicitly checking for overflow before doing this calculation, and aborting the GUI daemon in the event an overflow is detected.

gui-daemon/xside.c
ArrayBolt3: I took a look at all of the `process_xevent_*` functions, and none of them immediately stand out to be as being time-sensitive. I'll try to get around to testing it in practice soon (which should be done anyway since that will help shake out bugs if they're there).

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Overflow checking is now explicitly done and the GUI daemon is aborted *before* undefined behavior is caused if it is detected that this calculation would overflow.

DemiMarie:
For Wayland the tricky part is that the protocol is written assuming that events are dispatched in-order.  Buffering events therefore creates a risk of head-of-line blocking of events that really should not be delayed, like buffer release events indicating that memory can safely be reused.  Avoiding this requires reordering events, but that requires effort to ensure no bugs are introduced.

marmarek:
> For Wayland the tricky part is that the protocol is written assuming that events are dispatched in-order. Buffering events therefore creates a risk of head-of-line blocking of events that really should not be delayed, like buffer release events indicating that memory can safely be reused. Avoiding this requires reordering events, but that requires effort to ensure no bugs are introduced.

Reordering is problematic for X11 version too (see earlier comments). What events are problematic if delayed few hundreds of milliseconds? Will such delay of buffer release events cause practical issues beyond possibly requiring a bit more memory on the VM side?

ArrayBolt3:
I wouldn't expect delay to be a problem for X11 since, as I mentioned in one of the earlier comments, the X protocol is designed to operate over a network (for instance SSH with X tunneling). Networks incur a decent amount of latency. That latency causes delays similar to the ones this PR artificially introduces.

Wayland is a concern, however in practice waypipe exists and is used to provide network transparency to Wayland in a fashion similar to X11, and it appears to work well from what I've heard. That would cause the same latency and delay there, so if it's not a problem there it *probably* won't be a problem here.

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,69 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    size_t randsize;
+    union xbuf_rand xbuf_rand_data;
+
+    if (lower_bound >= g->xbuf_max_delay) {
+        fprintf(stderr,
+                "Bug detected - lower_bound >= g->xbuf_max_delay, events may get briefly stuck");
+        return g->xbuf_max_delay;
+    }
+
+    maxval = g->xbuf_max_delay - lower_bound + 1;
```
DemiMarie, line 2413: What is the `+ 1` for?  `g->xbuf_max_delay - lower_bound` is always going to be at least 1, so adding 1 again seems rather unusual.

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,69 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    } while (xbuf_rand_data.val >= UINT32_MAX - (UINT32_MAX % maxval));
+
+    randval = xbuf_rand_data.val % maxval;
+    return lower_bound + randval;
+}
+
+/* queue input event */
+static void xbuf_queue_xevent(Ghandles * g, XEvent xev)
+{
+    long current_time;
```
DemiMarie, line 2428: Why `long`?  It will always be equivalent to `int64_t`, but it would be clearer to be explicit here.

gui-daemon/xside.c
```diff
@@ -2449,6 +2509,40 @@ static void process_xevent(Ghandles * g)
...
+    while ((current_xbuf_entry = TAILQ_FIRST(&(g->xbuf_head)))
+        && (current_time >= current_xbuf_entry->time)) {
+        XEvent event_buffer = current_xbuf_entry->xev;
+        process_xevent_core(g, event_buffer);
+        TAILQ_REMOVE(&(g->xbuf_head), current_xbuf_entry, entries);
+        free(current_xbuf_entry);
+    }
+    current_xbuf_entry = TAILQ_FIRST(&(g->xbuf_head));
+    if (current_xbuf_entry == NULL) {
+        g->xbuf_next_timeout = 1000;
```
DemiMarie, line 2528: A named constant would be a bit nicer.

gui-daemon/xside.c
```diff
@@ -4242,6 +4336,18 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
             exit(1);
         }
     }
+
+    if ((setting =
+         config_setting_get_member(group, "xbuf_max_delay"))) {
+        int delay_val = config_setting_get_int(setting);
+        if (delay_val < 0 || delay_val > 5000) {
+            fprintf(stderr,
+                    "unsupported value ‘%d’ for xbuf_max_delay (must be >= 0 and <= 5000)",
```
DemiMarie, line 4345:
```suggestion
                    "unsupported value '%d' for xbuf_max_delay (must be >= 0 and <= 5000)",
```

Smart quotes (`‘` and `’`) are better typographical practice, but Qubes OS generally uses ASCII for anything that gets logged.  This is because various parts of the codebase clobber non-ASCII bytes for security reasons.

ArrayBolt3: what the heck, I typed that in Vim...? Not sure how that became smart quotes, but 100% agreed on the change, will integrate into the next iteration.

gui-daemon/xside.c
```diff
@@ -2386,53 +2400,120 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    if (maxval > UINT32_MAX)
+        return UINT32_MAX;
+    
+    do {
+        randval_raw = isaac_rand();
+    } while (randval_raw >= UINT64_MAX - (UINT64_MAX % maxval));
+
+    randval = (long)(randval_raw % maxval);
+    return lower + randval;
+}
```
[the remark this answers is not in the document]
marmarek: This comment is about older version

gui-daemon/xside.c
ArrayBolt3: The function is intended to return any value between `lower_bound` and `g->xbuf_max_delay` *inclusive*. Not adding a 1 here would result in `maxval = g->xbuf_max_delay` if `lower_bound = 0`. The modulo operator used in the line `randval = xbuf_rand_data.val % maxval` will return *at most* `maxval - 1`, so not adding 1 here would make this function return a number between `lower_bount` inclusive and `g->xbuf_max_delay` *exclusive*, which is not the intention here. The maximum amount of delay that can be inserted should be the amount of delay the user specifies, not the amount of delay the user specifies minus one millisecond.

gui-daemon/guid.conf
```diff
@@ -80,4 +80,10 @@ global: {
   # Timeout when waiting for qubes-gui-agent
   #
   # startup_timeout = 45;
+
+  # Maximum delay in milliseconds for X input buffering (used for obfuscating
+  # biometric data that could be obtained by observing keyboard and mosue
```
3hhh, line 85: mosue --> mouse

gui-daemon/guid.conf
```diff
@@ -80,4 +80,10 @@ global: {
...
+  # patterns). Set to 0 to disable buffering entirely.
+  #
+  # xbuf_max_delay = 0;
```
3hhh, line 88: If this can be ported to Wayland, I'd recommend to use an agnostic configuration name instead, e.g. `buf_max_delay` or `max_input_delay`. Also I'd then not mention Xorg in the comment.
This way users could easily migrate between the two display servers in the future.

gui-daemon/xside.c
ArrayBolt3: It looks like the existing code mostly uses `#define`s for this purpose, so I'm inclined to follow that pattern. If that's not acceptable, I'm happy to change it to a `const int`.

gui-daemon/xside.c
ArrayBolt3: I see what happened, the similar "unsupported value" line just a bit above also used smart quotes. I probably copy-pasted that and used it as a template. Both places will be fixed in the next iteration.

gui-daemon/guid.conf
ArrayBolt3: Renamed to "event buffering" and have switched to using `ebuf` throughout the code rather than `xbuf`. Will show up in the text iteration.

gui-daemon/guid.conf
ArrayBolt3: Fixed in next iteration.

gui-daemon/xside.c
ArrayBolt3: Fixed in next iteration.

gui-daemon/xside.c
ArrayBolt3: Created `VCHAN_DEFAULT_POLL_DURATION` define and put it both places where `wait_for_vchan_or_argfd_once` will eventually end up with the value.

ArrayBolt3:
Next iteration ready for review. Smoke-tested on Qubes OS R4.3, all requested changes implemented.

ArrayBolt3:
Had to force-push again because I forgot to update the commit message.

marmarek:
ebuf in code is fine, but for the setting name is rather cryptic. What about "events_delay_max" (or "events_max_delay"?) for the setting name? Or maybe something hinting at the purpose of the delay, like "input_cloak_max_delay"?

ArrayBolt3:
`events_max_delay` would work for me, I'll integrate that.

ArrayBolt3:
Switched to using `events_max_delay` terminology, also rebased onto the tip of main.

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+static uint32_t ebuf_random_delay(Ghandles * g, uint32_t lower_bound)
+{
+    uint32_t maxval;
+    uint32_t randval;
+    size_t randsize;
+    union ebuf_rand ebuf_rand_data;
+
+    if (lower_bound >= g->ebuf_max_delay) {
+        fprintf(stderr,
+                "Bug detected - lower_bound >= g->ebuf_max_delay, events may get briefly stuck");
```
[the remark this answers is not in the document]
ArrayBolt3: hmm, good point. Then perhaps the `>=` should change to a `>`?

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+/* queue input event */
+static void ebuf_queue_xevent(Ghandles * g, XEvent xev)
+{
+    int64_t current_time;
+    uint32_t random_delay;
+    struct ebuf_entry *new_ebuf_entry;
+    uint32_t lower_bound;
+
+    current_time = ebuf_current_time_ms();
+    lower_bound = min(max(g->ebuf_prev_release_time - current_time, 0), g->ebuf_max_delay);
```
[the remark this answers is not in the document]
ArrayBolt3: If you don't, you end up with out-of-order release timestamps. The release times are stored in each `ebuf_entry` as an absolute timestamp, which is generated by taking an arrival timestamp (`current_time`) and adding a delay to it to get a release timestamp. Each event is only released once two conditions are met:

* The event is at the head of the event queue, and
* The current timestamp at time of check is equal to or greater than the scheduled release timestamp.

This implies that each subsequent delay *must* result in a release timestamp greater than or equal to the previous release timestamp. Otherwise you end up with events later in the queue having a scheduled release time *earlier* than events earlier in the queue, meaning they will get stuck longer than they were scheduled to and will end up being released all at once as soon as all "clogging" events ahead of them have been released.

In order to always generate equal or larger timestamps than those generated previously, it is necessary to ensure that `current_time + random_delay >= g->ebuf_prev_release_time`. This requires that `random_delay` be no less than `g->ebuf_prev_release_time - current_time`. That is what the line of code in question does. It also includes some sanity-ensuring boundaries to make sure that `lower_bound` is never less than zero or greater than `g->ebuf_max_delay`.

Should this be included in a comment?

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+        return g->ebuf_max_delay;
+    }
+
+    maxval = g->ebuf_max_delay - lower_bound + 1;
+    
+    do {
+        randsize = getrandom(ebuf_rand_data.raw, sizeof(uint32_t), 0);
+        if (randsize != sizeof(uint32_t))
+            continue;
+    } while (ebuf_rand_data.val >= UINT32_MAX - (UINT32_MAX % maxval));
```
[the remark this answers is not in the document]
ArrayBolt3: Hmm, that's true. I'm not totally sure how to solve that in a simple way. Suggestions?

gui-daemon/xside.c
```diff
@@ -4364,11 +4472,13 @@ int main(int argc, char **argv)
...
     parse_cmdline(&ghandles, argc, argv);
     get_boot_lock(ghandles.domid);
+    /* init event queue */
+    TAILQ_INIT(&(ghandles.ebuf_head));
 
     if (!ghandles.nofork) {
         // daemonize...
         if (pipe(pipe_notify) < 0) {
-            perror("canot create pipe:");
+            perror("cannot create pipe:");
```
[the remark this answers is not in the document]
ArrayBolt3: Will be pushed as a separate commit in next iteration.

gui-daemon/xside.c
```diff
@@ -4237,11 +4333,23 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
             g->disable_override_redirect = 0;
         else {
             fprintf(stderr,
-                    "unsupported value ‘%s’ for override_redirect (must be ‘disabled’ or ‘allow’\n",
+                    "unsupported value '%s' for override_redirect (must be 'disabled' or 'allow')\n",
```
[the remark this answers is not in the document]
ArrayBolt3: Will be pushed as a separate commit in next iteration.

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+{
+    int64_t timeval;
+    struct timespec spec;
+    clock_gettime(CLOCK_MONOTONIC, &spec);
+    timeval = (((int64_t)spec.tv_sec) * 1000LL) + (((int64_t)spec.tv_nsec) / 1000000LL);
+    return timeval;
+}
+
+/* get random delay value */
+static uint32_t ebuf_random_delay(Ghandles * g, uint32_t lower_bound)
```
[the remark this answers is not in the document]
ArrayBolt3: Rewritten to be self-contained.

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    union ebuf_rand ebuf_rand_data;
+
+    if (lower_bound >= g->ebuf_max_delay) {
+        fprintf(stderr,
+                "Bug detected - lower_bound >= g->ebuf_max_delay, events may get briefly stuck");
+        return g->ebuf_max_delay;
+    }
+
+    maxval = g->ebuf_max_delay - lower_bound + 1;
+
```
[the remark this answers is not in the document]
ArrayBolt3: Found it, removed, thanks for catching that!

gui-daemon/xside.c
```diff
@@ -2386,11 +2388,71 @@ static void process_xevent_xembed(Ghandles * g, const XClientMessageEvent * ev)
...
+    current_time = ebuf_current_time_ms();
+    lower_bound = min(max(g->ebuf_prev_release_time - current_time, 0), g->ebuf_max_delay);
+    random_delay = ebuf_random_delay(g, lower_bound);
+    new_ebuf_entry = malloc(sizeof(struct ebuf_entry));
+    if (new_ebuf_entry == NULL) {
+        perror("Could not allocate ebuf_entry:");
+        exit(1);
+    }
+    if (current_time > 0 && random_delay > (LONG_MAX - current_time)) {
```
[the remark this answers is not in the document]
ArrayBolt3: Suggestion added as shown, thanks!

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: "Fixed" - the error message is now logged only if `lower_bound > upper_bound`. This particular code block will still be triggered for the `lower_bound = upper_bound` case, but it's not actually a bug and so no error message will be logged, it will just return `upper_bound`. This is, to my awareness, the correct behavior in this scenario.

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Explanatory comment added.

gui-daemon/xside.c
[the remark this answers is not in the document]
ArrayBolt3: Thought about it for a while, the solution I came up with was to change the `while` bit to this:

```c
while (ebuf_rand_data.val > UINT32_MAX - ((uint32_t)((((uint64_t)UINT32_MAX) + 1) % maxval)));
```

It's... ugly... but it should work I believe. Critically, if `maxval` is a multiple of 2\*\*32, the modulo operation will now return 0, meaning that in order to flunk the test `ebuf_rand_data.val` would have to be greater than `UINT32_MAX`. But `ebuf_rand_data.val` is a `uint32_t` so that's impossible, thus any value will be accepted. On the other hand, if `maxval` is not a multiple of 2\*\*32, the modulo operation will return the number of "problem" values at the top of the `uint32_t` range that need to be avoided. If `ebuf_rand_data.val` is one of those problem values, a different random value will be tried.

I've thought this through with 3-bit integers and I believe it works, however I've not proven it. It would probably be a good idea to test this thoroughly with 8-bit integers, and ensure no modulo bias is visible.

ArrayBolt3:
Should be ready for another review. Ended up doing two force-pushes because I forgot to sign my commit the first time.

ArrayBolt3:
Gentle ping, is there any progress here, or any further changes that need to be made?

marmarek:
I've tested it a bit and have an observation: the more events are queued (or rather: the quicker new events are produced), the less random delay is. You can easily get into this situation with mouse move events. After initial short time, further events get delay near the max configured value (in the top 10-20ms, depending on movement speed). Then, you need to wait the configured max delay to regain full random delay range.
In practice, it means the random range is limited by the frequency of incoming events (except short initial time). 

This is probably not an issue for keyboard events (delay between those events in practice is large enough to leave some room for randomness). But may be an issue for mouse events.
In practice this is okay, right? I think you mostly care about obfuscating typing pattern, which should still be okay (except short time after moving the mouse).
This also means that setting higher values for the max delay has not only UX implications, but also affects how long you need to wait after moving the mouse to regain full randomness range.

But if you do care about obfuscating mouse movements too, this PR isn't enough in its current shape. I haven't read research on this, but I'd expect that profiling based on mouse movements isn't only about speed but also (mostly) about move patterns (directions, mouse travel etc). This you don't change with delays at all. Some simple approach could be dropping some move events (which incidentally would also help with the earlier issue), but it's a risky thing to do (knowing which moves are safe to drop to not break functionality).

Anyway, I'm okay with merging this as is. But whoever wants to use this feature, needs to be aware of the limitation, and be careful about choosing max delay value.

marmarek (approved):

marmarek:
> Unless there's a good reason for the current behavior I think something like `max(last, now) + random(min, max)` would be most likely better.

The problem with this is you get no upper bound on the actual delay then. Imagine you set max delay to 500ms and then move the mouse a bit which, say, generate 200 events. Now you need to wait on average for 5s for the queue to process. It isn't hard to imagine more extreme cases...

ArrayBolt3:
To be clear, the issue with "shrinking timeouts" exists in the upstream kloak code which this PR was based upon, and it still did a good job at obfuscating user identity according to tests run by kloak's original author. (Sadly the tool he used, KeyTrac, is no longer publicly available, so it's not so easy to test this anymore.) Thus I would say that this *is* intended behavior, since it's in line with behavior that is known to work well enough. (That being said, I do see how mouse and keyboard events will end up conflicting with each other to some degree, i.e. if someone's typing and moving the mouse at the same time, it will *really* reduce or eliminate the typing randomization. That's still in line with what kloak does, but it probably hasn't been tested the same way as a keyboard-only use case.)

Separating mouse and keyboard events is theoretically doable, and I'm willing to give it a shot if it would be useful. However, this does mean that the user may be moving the mouse, clicking on things, etc., while the keyboard buffer is still being flushed. With a larger timeout, I can easily see this resulting in the user's keystrokes going to the wrong app or input widgets.

ArrayBolt3:
I've been working on a new implementation of kloak for non-Qubes systems that fixes the mouse anonymization issues of the original kloak implementation. The algorithm *should* be able to be ported to Qubes OS. The code is at https://github.com/ArrayBolt3/kloak-v2 if you're interested, the way the algorithm works is that when mouse movements are made, the last queued event is checked to see if it is a mouse event, and if so, that event's intended end location for the mouse are changed rather than adding a new mouse event to the queue. It's currently designed specifically for wlroots-based Wayland compositors (i.e. will not work on X11), so if you want to test it you'll probably want to test it using a fairly recent version of labwc (which is what I tested against). The Qubes reimplementation won't have the problem of being X11-incompatible since it will be working specifically with X11 input events rather than libinput events.

3hhh:
On 5/5/25 03:24, Aaron Rainbolt wrote:
> ArrayBolt3 left a comment (QubesOS/qubes-gui-daemon#149)
> 
> I've been working on a new implementation of kloak for non-Qubes systems that fixes the mouse anonymization issues of the original kloak implementation. The algorithm *should* be able to be ported to Qubes OS. The code is at https://github.com/ArrayBolt3/kloak-v2 if you're interested, the way the algorithm works is that when mouse movements are made, the last queued event is checked to see if it is a mouse event, and if so, that event's intended end location for the mouse are changed rather than adding a new mouse event to the queue.

Hm... Isn't a jumping mouse unique for tracking algorithms as well?
Yes, it reduces the number of input data, but as long as the number of users of such an algorithm remains low, it is an easily discernible data point imho.

The other ideas which were recently discussed on the forums as "[Qubes Forum] How are qubes users dealing with mouse movement fingerprinting? [All around Qubes]" - unfortunately in the private "All around Qubes" section - (e.g. https://github.com/r3t4k3r/mouser/ & https://github.com/rvaiya/warpd) seem to wait for the user to make a final mouse positioning decision before passing the mouse position to the OS. That's pretty similar to your approach and should therefore be discernible, too.

Of course if user adoption grows, each such user will just be identifiable as a member of that growing group - at least based on that metric.

ArrayBolt3:
@3hhh Yes, it will, but we have an identical situation with the way keyboard patterns are obfuscated by kloak. The only way to make an obfuscation program for keyboard or mouse activity that could prevent fingerprinting without requiring a substantial userbase would be to clone someone *else's* keyboard and mouse usage patterns, which is many times more complicated if it's even possible with existing technology. kloak in its current incarnation is enabled by default in Whonix though, and the new version of kloak will likely be shipped with and enabled by default in Whonix as well, which will provide the needed userbase to allow for this to actually provide enhanced anonymity.

adrelanos:
Quote [Whonix: Identifiers Design Goals](https://www.whonix.org/wiki/Protocol-Leak-Protection_and_Fingerprinting-Protection#Identifiers_Design_Goals):

> Should the goal be,
> 
> A) a shared personality: to have all Whonix appear with the same uniform fingerprint at all times, OR
> B) a virtual personality: to invent and emulate a different unique fingerprint for each user?

> Whonix design, at the time of writing, is A) (shared personality).

> Tor and the Tor Browser take a similar approach to this issue. They do not try to generate a new random identity (pseudonym) each time they are used. Instead, their strategy is to make all users appear the same to outside observers. The Tor Project refers to this concept as Anonymity Loves Company (you can search the web for this term for more background). Since Whonix is designed as an extension of Tor, it follows the same principle.

----

Emulating a different virtual personality for each user would require extensive research, complex algorithms or even [artificial intelligence (AI)](https://www.kicksecure.com/wiki/Artificial_intelligence).

Due to piratically unsolvable issues like [VM Fingerprinting](https://www.whonix.org/wiki/VM_Fingerprinting) and [Browser Fingerprinting](https://www.whonix.org/wiki/Data_Collection_Techniques#Browser_Fingerprinting), we don't attempt to implement a virtual personality and stick with the shared personality design.

adrelanos:
Documented here just now:
[Keystroke and Mouse Deanonymization](https://www.whonix.org/wiki/Keystroke_and_Mouse_Deanonymization) wiki page, limitations, [Shared Personality](https://www.whonix.org/wiki/Keystroke_and_Mouse_Deanonymization#Shared_Personality).

3hhh:
I see. Looks like we should then align on an approach with a large enough user base for the mouse movement issue.

If the mouse will be jumping anyway, I'd probably wait for the final click to be made before passing the position information to the VM. That's the least amount of information which can be passed to an attacker. dom0 can still draw the mouse while moving anyway so that the mouse position is always clear to the user.

Is there an enhancement issue for this open already or should we open one?

marmarek:
> wait for the final click to be made before passing the position

There a significant issue with this approach - you can't have hover effect anymore. For example when you want to see a tooltip, or an URL before clicking it.

ArrayBolt3:
Yeah, that's why I went with automatic updates of the mouse position. Delaying movement events until a click is made causes all sorts of problems - hover is broken, as Marek mentioned, but drag will also be somewhat broken, as well as anything that actually needs to track mouse movement to provide functionality (this is kind of a silly example but slither.io is a game that needs to track mouse movements to even be playable). Mouser tries to work around this by allowing the use of a middle click to move the mouse cursor, but that only solves the hover problem, and it also means the middle mouse button (which some applications have a legitimate use for) will also become unusable.

PR #150, opened 2024-10-06
Make maximum clipboard buffer lenght configurable

alimirjamali:
1st part of fix. Making the clipboard buffer size configurable via `max_clipboard_size` setting within `guid.conf`

fixes: https://github.com/QubesOS/qubes-issues/issues/9296

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -859,60 +859,83 @@ static void save_clipboard_metadata(struct clipboard_metadata *metadata) {
     umask(old_umask);
 }
 
-static bool load_clipboard_metadata(struct clipboard_metadata *metadata) {
+static bool load_clipboard_metadata(struct clipboard_metadata *metadata, bool logging) {
     FILE *file;
     char line[256];
-    char key[256];
-    char value[256];
+    char key[256] = {0};
+    char value[256] = {0};
 
     file = fopen(QUBES_CLIPBOARD_FILENAME ".metadata", "r");
     if (!file) {
-        perror("Can not open " QUBES_CLIPBOARD_FILENAME ".metadata file");
+        if (logging)
+            perror("Can not open " QUBES_CLIPBOARD_FILENAME ".metadata file");
         return false;
     }
     // Load JSON format
     while (fgets(line, sizeof(line), file) != NULL) {
+        if (strlen(line) == 0) continue;
+        if (line[strlen(line) - 1] != '\n') {
+            fprintf (stderr, "Clipboard metadata line over 255 characters: %s...\n", line);
+            return false;
+        }
+        line[strlen(line) - 1] = 0;   // remove the trailing `\n` for easy parsing
         if (strcmp(line, "") == 0) continue;
         if (strcmp(line, "{") == 0) continue;
         if (strcmp(line, "}") == 0) continue;
-        if (! sscanf(line, "\"%[A-Za-z0-9]\":%[\"A-Za-z0-9]s,", key, value)) {
-            fprintf (stderr, "Failed to parse metadata line: %s\n", line);
+        if (sscanf(line, "\"%[A-Za-z0-9_-]\":%[\"A-Za-z0-9_-]", key, value) != 2) {
+            fprintf (stderr, "Failed to parse clipboard metadata line: %s\n", line);
             return false;
         }
         if (strcmp(key, "vmname") == 0) {
             /* value should be less than allowed maximum vmlenght + 2
              * considering the quotation marks */
-            if (strlen(value) >= 32 + 2) return false;
-            if (strlen(value) < 2) return false;
+            int vlen = strlen(value);
+            if ((vlen >= 32 + 2) ||
+                (vlen < 2) ||
+                (value[0] != '"') ||
+                (value[vlen - 1] != '"')) {
+                    fprintf (stderr, "Clipboard vmname value should be less than 32 characters and between double-quotes: %s\n", line);
+                    return false;
+            }
             strncpy (metadata->vmname, value + 1, strlen(value) - 2);
-        } else if (strcmp(key, "xevent_timestamp") == 0) {
-            sscanf(value, "%lu", &metadata->xevent_timestamp);
+            continue;
+        }
+
+        /* From this point on, all recognized values are essentially integers */
+        long unsigned int dummy;
+        if (sscanf(value, "%lu", &dummy) != 1) {
+            fprintf (stderr, "Failed to parse clipboard metadata: key=%s, value=%s\n", key, value);
+            return false;
+        }
+
+        if (strcmp(key, "xevent_timestamp") == 0) {
+            metadata->xevent_timestamp = dummy;
         } else if (strcmp(key, "successful") == 0) {
-            sscanf(value, "%d", (int *)&metadata->successful);
+            metadata->successful = dummy;
         } else if (strcmp(key, "copy_action") == 0) {
-            sscanf(value, "%d", (int *)&metadata->copy_action);
+            metadata->copy_action = dummy;
         } else if (strcmp(key, "paste_action") == 0) {
-            sscanf(value, "%d", (int *)&metadata->paste_action);
+            metadata->paste_action = dummy;
         } else if (strcmp(key, "malformed_request") == 0) {
-            sscanf(value, "%d", (int *)&metadata->malformed_request);
+            metadata->malformed_request = dummy;
         } else if (strcmp(key, "oversized_request") == 0) {
-            sscanf(value, "%d", (int *)&metadata->oversized_request);
+            metadata->oversized_request = dummy;
         } else if (strcmp(key, "cleared") == 0) {
-            sscanf(value, "%d", (int *)&metadata->cleared);
+            metadata->cleared = dummy;
         } else if (strcmp(key, "qrexec_clipboard") == 0) {
-            sscanf(value, "%d", (int *)&metadata->qrexec_clipboard);
+            metadata->qrexec_clipboard = dummy;
         } else if (strcmp(key, "sent_size") == 0) {
-            sscanf(value, "%d", &metadata->sent_size);
+            metadata->sent_size = dummy;
         } else if (strcmp(key, "buffer_size") == 0) {
-            sscanf(value, "%d", &metadata->buffer_size);
+            metadata->buffer_size = dummy;
         } else if (strcmp(key, "protocol_version_vmside") == 0) {
-            sscanf(value, "%x", &metadata->protocol_version_vmside);
+            metadata->protocol_version_vmside = dummy;
         } else if (strcmp(key, "protocol_version_xside") == 0) {
-            sscanf(value, "%x", &metadata->protocol_version_xside);
+            metadata->protocol_version_xside = dummy;
         }
     }
     fclose(file);
-    return metadata->cleared;
+    return true;
 }
 
 /* caller must take inter_appviewer_lock first */
@@ -926,10 +949,29 @@ static void clear_clipboard(struct clipboard_metadata *metadata) {
     save_clipboard_source_vmname("");
 }
 
-static Time get_clipboard_xevent_timestamp() {
+/* caller must take inter_appviewer_lock first */
+static Time get_clipboard_xevent_timestamp(bool logging) {
     struct clipboard_metadata metadata = {0};
-    load_clipboard_metadata(&metadata);
-    return metadata.xevent_timestamp;
+    FILE *file;
+
+    /* do not do a detailed logging for metadata parsing. handle non-existent
+     * .metadata file independently here */
+    file = fopen(QUBES_CLIPBOARD_FILENAME ".metadata", "r");
+    if (!file) {
+        if (logging)
+            perror("Can not get xevent timestamp from non-existent " QUBES_CLIPBOARD_FILENAME ".metadata");
+        return 0;
+    } else {
+       fclose(file);
+    }
+
+    if (!load_clipboard_metadata(&metadata, logging)) {
+        if (logging)
+            perror("Can not get xevent timestamp from " QUBES_CLIPBOARD_FILENAME ".metadata");
+        return 0;
+    } else {
+        return metadata.xevent_timestamp;
+    }
 }
 
 /* fetch clippboard content from file */
@@ -1193,8 +1235,15 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
         clear_clipboard(&metadata);
         return;
     }
+    if (metadata.sent_size == 0) {
+        /* source vm clipboard is empty. clear .data and update .metadata */
+        free(untrusted_data);
+        metadata.successful = true;
+        clear_clipboard(&metadata);
+        return;
+    }
     inter_appviewer_lock(g, 1);
-    clipboard_file_xevent_time = get_clipboard_xevent_timestamp();
+    clipboard_file_xevent_time = get_clipboard_xevent_timestamp(g->log_level > 0);
     /* X11 time is just 32-bit miliseconds counter, which make it wrap every
      * ~50 days - something that is realistic. Handle that wrapping too. */
     if (clipboard_file_xevent_time - g->clipboard_xevent_time < (1UL<<31)) {
@@ -1469,7 +1518,7 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
         if (ev->type != KeyPress)
             return 1;
         inter_appviewer_lock(g, 1);
-        clipboard_file_xevent_time = get_clipboard_xevent_timestamp();
+        clipboard_file_xevent_time = get_clipboard_xevent_timestamp(g->log_level > 0);
         /* X11 time is just 32-bit miliseconds counter, which make it wrap every
          * ~50 days - something that is realistic. Handle that wrapping too. */
         if (clipboard_file_xevent_time - ev->time < (1UL<<31)) {
@@ -3934,6 +3983,7 @@ struct option longopts[] = {
     { "title-name", no_argument, NULL, 'T' },
     { "trayicon-mode", required_argument, NULL, opt_trayicon_mode },
     { "screensaver-name", required_argument, NULL, opt_screensaver_name },
+    { "max-clipboard-size", required_argument, NULL, 'X' },
     { "help", no_argument, NULL, 'h' },
     { "version", no_argument, NULL, 'V' },   // V is virtual and not a short option
     { 0, 0, 0, 0 },
@@ -4243,6 +4293,16 @@ static void parse_cmdline(Ghandles * g, int argc, char **argv)
             }
             g->screensaver_names[screensaver_name_num++] = strdup(optarg);
             break;
+        case 'X':
+            unsigned int value;
+            if (sscanf(optarg, "%u", &value) != 1)
+                errx(1, "maximum clipboardsize '%s' is not an integer", optarg);
+            if (value > MAX_CLIPBOARD_BUFFER_SIZE || value < MIN_CLIPBOARD_BUFFER_SIZE)
+                errx(1, "unsupported value ‘%d’ for --max-clipboard-size "
+                        "(must be between %d to %d characters)",
+                        value, MAX_CLIPBOARD_BUFFER_SIZE, MIN_CLIPBOARD_BUFFER_SIZE);
+            g->clipboard_buffer_size = value;
+            break;
         default:
             usage(stderr);
             exit(1);
@@ -4365,7 +4425,6 @@ static void parse_vm_config(Ghandles * g, config_setting_t * group)
                     "unsupported value ‘%d’ for max_clipboard_size "
                     "(must be between %d to %d characters).\n",
                     value, MAX_CLIPBOARD_BUFFER_SIZE, MIN_CLIPBOARD_BUFFER_SIZE);
-            exit(1);
         } else {
             g->clipboard_buffer_size = value;
         }
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -69,7 +69,7 @@
 /* Supported protocol version */
 
 #define PROTOCOL_VERSION_MAJOR UINT32_C(1)
-#define PROTOCOL_VERSION_MINOR UINT32_C(7)
+#define PROTOCOL_VERSION_MINOR UINT32_C(8)
 #define PROTOCOL_VERSION(x, y) ((x) << 16 | (y))
 
 #if !(PROTOCOL_VERSION_MAJOR == QUBES_GUID_PROTOCOL_VERSION_MAJOR && \
@@ -791,25 +791,7 @@ static struct windowdata *check_nonmanaged_window(Ghandles * g, XID id)
     return item->data;
 }
 
-/* caller must take inter_appviewer_lock first */
-static Time get_clipboard_file_xevent_timestamp() {
-    FILE *file;
-    Time timestamp;
-
-    file = fopen(QUBES_CLIPBOARD_FILENAME ".xevent", "r");
-    if (!file) {
-        perror("open " QUBES_CLIPBOARD_FILENAME ".xevent");
-        return 0;
-    }
-    if (fscanf(file, "%lu", &timestamp) < 1) {
-        fprintf(stderr, "Failed to load " QUBES_CLIPBOARD_FILENAME " file (empty?)\n");
-        timestamp = 0;
-    }
-    fclose(file);
-    return timestamp;
-}
-
-/* caller must take inter_appviewer_lock first */
+/* caller must take inter_appviewer_lock first. for legacy applications */
 static void save_clipboard_file_xevent_timestamp(Time timestamp) {
     FILE *file;
     mode_t old_umask;
@@ -826,6 +808,7 @@ static void save_clipboard_file_xevent_timestamp(Time timestamp) {
     umask(old_umask);
 }
 
+/* for legacy applications */
 static void save_clipboard_source_vmname(const char *vmname) {
     FILE *file;
     mode_t old_umask;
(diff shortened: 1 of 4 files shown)


gui-daemon/xside.c
```diff
@@ -1019,7 +1019,7 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     if (g->log_level > 0)
         fprintf(stderr, "handle_clipboard_data, len=0x%x\n",
             untrusted_len);
-    if (untrusted_len > MAX_CLIPBOARD_SIZE) {
+    if (untrusted_len > g->clipboard_buffer_size) {
```
marmarek, line 1022: This is not that simple. As you can see here, if VM provides too much data, it gets disconnected (`exit(1)` below). So, there either needs to be a mechanism to tell the max size to the VM, or a more graceful handling of too long data. For the latter it needs to be read from the client (read_data) in full, even if too long, otherwise remainder would be interpreted as subsequent commands. And be careful to not turn this into denial-of-service by allocating VM-controlled arbitrary large memory buffer below (when reading to discard, do it in chunks).

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: Will work on it. If this is ever going to be user configurable, maybe it is better to keep it simple for the time and set `MAX_CLIPBOARD_BUFFER_SIZE` in PR to 250000 characters. This was even accepted before the [Big Requests Extension](https://www.x.org/releases/X11R7.7/doc/bigreqsproto/bigreq.html) and INCR in mid 90s. 

User's major concern is truncation without notification. It would better to spend more time on that.

Unrelated question. Is the Gitlab CI/CD infra still broken? All the Fedora & Debian builds have failed.

marmarek: > Unrelated question. Is the Gitlab CI/CD infra still broken? All the Fedora & Debian builds have failed.

It looks like viewing results without logging into gitlab.com is still broken. But the reason for the failures is that this PR really fails to build.

marmarek: Oh, and technically, when bumping maximum clipboard size, it needs also bumping protocol version, so the agent sees that it can send more data.

alimirjamali: > But the reason for the failures is that this PR really fails to build.

I tried to cheat a little since setting up the build environment on Fedora 41 is problematic (and it appears that CI/CD also lacks Fedora 41 builds). But no worries. It is finally the moment for me to learn builderv2 instead of quickly testing everything locally.

alimirjamali: My overall plan & design is going to be simple. Letting agent to send up to the new `MAX_CLIPBOARD_BUFFER_SIZE + 1` characters (250KB+1) in [this line](https://github.com/QubesOS/qubes-gui-agent-linux/blob/main/gui-agent/vmside.c#L990). Then `xside` should simply reject sizes over `g->clipboard_buffer_size` and clear the buffer [in this block](https://github.com/QubesOS/qubes-gui-daemon/blob/main/gui-daemon/xside.c#L1022-L1026); with proper notification to user about it.

The wasted 250KB for this simple mechanism is not that bad and is comparable to some 256x256 pixel PNGs. It is better than sending a truncated clipboard IMO. Hopefully this should require minimum modification to agent side and fully backward compatible (templates with old protocol will truncate the clipboard as before).

This will need more testing and changes to 3 repositories. I am currently doing my testings in Fedora 40.

gui-daemon/xside.c
```diff
@@ -1013,16 +1013,17 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
     if (g->log_level > 0)
         fprintf(stderr, "handle_clipboard_data, len=0x%x\n",
             untrusted_len);
-    if (untrusted_len > MAX_CLIPBOARD_SIZE) {
-        fprintf(stderr, "clipboard data len 0x%x?\n",
+    if (untrusted_len > g->clipboard_buffer_size) {
+        clipboard_buffer_exceeded = true;
+        fprintf(stderr, "clipboard data len 0x%x exceeds maximum allowed!\n",
             untrusted_len);
-        exit(1);
```
marmarek, line 1025: This needs more care. If VM gave some large number, you can't simply pass that to malloc below. That would allow VM to make gui-daemon allocate up to 4GB memory, most likely crashing it and starving few other processes too.
If limit is exceeded, you need to read and discard that data in chunks (that are not overly big). Or, apply some other upper limit (in the description you said it will be at most MAX_CLIPBOARD_SIZE+1, so maybe exit if that limit is exceeded too?).

gui-daemon/xside.c
```diff
@@ -1057,7 +1058,8 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
         show_error_message(g, "secure copy: failed to open file " QUBES_CLIPBOARD_FILENAME);
         goto error;
     }
-    if (fwrite(untrusted_data, 1, untrusted_data_sz, file) != untrusted_data_sz) {
+    if (! clipboard_buffer_exceeded)    // do not write anything to output if size exceeds limit
```
marmarek, line 1061: I don't think empty file is a good idea for signaling that limit was exceeded. This may happen for other reasons too (for example when really VM clipboard was empty). Either adjust the message in https://github.com/QubesOS/qubes-desktop-linux-manager/pull/223 that too much data is just one of possible reasons, or have some other method to inform about too much data (yet another file?).

gui-daemon/xside.c
```diff
@@ -3842,6 +3847,11 @@ static void parse_cmdline_config_path(Ghandles * g, int argc, char **argv)
         } else if (opt == 'h') {
             usage(stdout);
             exit(0);
+        } else if (opt == 'V') {
+            fprintf(stdout, "Qubes GUI Daemon v%d.%d\n",
+                    PROTOCOL_VERSION_MAJOR,
+                    PROTOCOL_VERSION_MINOR);
```
marmarek, line 3853: This is not a program version, but protocol version.

gui-daemon/xside.c
```diff
@@ -3807,6 +3810,8 @@ static void usage(FILE *stream)
     fprintf(stream, " --title-name, -T\tprefix window titles with VM name\n");
     fprintf(stream, " --trayicon-mode\ttrayicon coloring mode (see below); default: tint\n");
     fprintf(stream, " --screensaver-name\tscreensaver window name, can be repeated, default: xscreensaver\n");
+    fprintf(stream, " --help, -h\tshow command help\n");
+    fprintf(stream, " --version\tshow program version\n");
```
marmarek, line 3814:
```suggestion
    fprintf(stream, " --version\tshow protocol version\n");
```
?

gui-daemon/xside.h
```diff
@@ -90,6 +90,11 @@
 #include <qubes-gui-protocol.h>
 #include "util.h"
 
+/* default, minimum and maximum clipboard buffer lenght */
+#define DEFAULT_CLIPBOARD_BUFFER_SIZE 64000   // make this 1000 bytes less than previous value so it works with older vmside agent
```
marmarek, line 94: How setting lower default limit makes it work with the old agent? It will reject sizes that were allowed before. If you want to detect (likely) truncation with the old agent, then maybe set it to 64999? It isn't 100% correct (if user really tries to copy exactly 65000 chars, it will be rejected now), but may be close enough. Or keep it at 65000 so the old agent is not affected at all.

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: Yes. Right. It would be the case of a malicious VM that does not comply with the vmside patch. It is possible and should be handled. I will do it.

gui-daemon/xside.c
alimirjamali: I also thought about it. Yet another file or maybe adding some metadata to `qubes-clipboard.bin.source`. The former option sounds more appealing. BTW, I think it would be better to keep it a structured file this time. So it could be used later for other purposes as well (such as adding some metadata).

gui-daemon/xside.c
alimirjamali: I thought it is good to have. Should I update it to print the protocol version or remove it entirely?

gui-daemon/xside.c
alimirjamali: OK

gui-daemon/xside.h
alimirjamali: This will work (if vmside protocol version is below 1.8). Anything above 64999 means either the vmside actual size was exactly 65000 characters (very low probability) or over it. Overall, rejecting sizes over 64999 characters (by default) is a good compromise if the user has a template with older vmside.

This default could be readjusted to 256000 after all templates with vmside versions of 1.7 or older are EOL.

I will adjust to 64999

alimirjamali: I hope this could be reconsidered. I made the new default 64000 characters since it is kinda confusing to explain to users that the new default is 1 characters less than the previous default (64999). So the xside protocol version 1.8 could detect clipboard buffer exceeding on the vmsides protocol version less than 1.8.

Please advise.

marmarek: Maybe make default 65000, but on the old protocol lower it to 64999 ? The limit can't be above 65000 on the old protocol anyway.

gui-daemon/xside.c
alimirjamali: I am working on `save-clipboard-metadata` function. Technically the output `.metadata` file could be used to phase-out `.source` and `.xevent` files in future Qubes OS releases. I will make it a new commit for this PR instead of amending it. So it would be easier to follow-up and review.

gui-daemon/xside.c
```diff
@@ -842,6 +842,31 @@ static void save_clipboard_source_vmname(const char *vmname) {
...
+    fprintf(file, "clipboard-buffer-size=%d\n", g->clipboard_buffer_size);
+    fprintf(file, "gui-protocol-version-xside=%d.%d\n",
+        PROTOCOL_VERSION_MAJOR, PROTOCOL_VERSION_MINOR);
+    fprintf(file, "gui-protocol-version-vmside=%d.%d\n",
+        g->protocol_version >> 16, g->protocol_version & 0xFFFF);
+    // other key,value pairs could be added if needed in future
+    fclose(file);
+    umask(old_umask);
+}
+
```
marmarek, line 869: It should not only write the file, but also read it and use in relevant places, if not older than the data file.

alimirjamali: > It should not only write the file, but also read it and use in relevant places, if not older than the data file.

My original intention was to use it only for Global Clipboard systray widget. But this `.metadata` file could technically replace the old .source and .xevent files entirely. Beside the gui-daemon and systray widget, should I look elsewhere to fix? Is there anything else using the .source and .xevent files?

gui-daemon/xside.c
alimirjamali: This one is done with more care. It will `exit(1)` if vm sends over `MAX_CLIPBOARD_SIZE` (i.e. 65000) on Protocol v1.7 and over `MAX_CLIPBOARD_BUFFER_SIZE + 1` (i.e. 320001) on protocol 1.8 or newer.

I will add INCR detection here to reject INCR entirely.

gui-daemon/xside.c
marmarek: > My original intention was to use it only for Global Clipboard systray widget.

Intentional duplication of info (not only in transitional period) is not a good idea. The comment above suggests you already planned migration.

> Beside the gui-daemon and systray widget, should I look elsewhere to fix? Is there anything else using the .source and .xevent files?

There are also buttons "copy to global clipboard" in a bunch of places:
- qube logs in qube manager
- "this device" tab in qubes global config
- updater gui

alimirjamali: > > My original intention was to use it only for Global Clipboard systray widget.
> 
> Intentional duplication of info (not only in transitional period) is not a good idea. The comment above suggests you already planned migration.
> 
> > Beside the gui-daemon and systray widget, should I look elsewhere to fix? Is there anything else using the .source and .xevent files?
> 
> There are also buttons "copy to global clipboard" in a bunch of places:
> 
>     * qube logs in qube manager
>     * "this device" tab in qubes global config
>     * updater gui


This is almost ready for review. I am performing more local tests to assure it is working well. I decided to switch to JSON for `.metadata`. And completely phased-out `.source` and `.xevent` within the `qubes-guid` source. I have to work on the GTK and Qt tools next if this one is OK. I think the Qt and GTK tools should handle both the previous format and the new one for a while (if this is approved).

marmarek: Old files at least needs to be still created, for applications (possibly even 3rd-party ones) that still use those to get info about the clipboard.

alimirjamali: I restored creation of `.source` and `.xevent` files. I will work on the clipboard widget next.

The CI/CD build of this and `gui-agent` fail since they depends on the pending changes to headers of `gui-common`. My local build of qubes-guid takes almost 5 minutes. I guess if Demi's work to transition GUIVM to Wayland is going to take a long of time, it might be good idea to consider breaking `xside.c` to smaller parts (currently over 4700 lines).  For example clipboard functions in one file, keyboard processing in one file, ...

alimirjamali: This is now ready for review @marmarek 

I converted all 5 connected PRs from draft to ready.

alimirjamali:
PipelineRetryFailed

marmarek:
PipelineRetryFailed

marmarek:
PipelineRetry

marmarek:
Pasting doesn't work:
```
Failed to parse metadata line: {
```
But that's probably because copying failed:
```
Can not open /var/run/qubes/qubes-clipboard.bin.metadata file: No such file or directory
```

marmarek:
Ok, the message on copy is not really a failure - the first one is expected to not have this file. Can you suppress the error message in this case?

gui-daemon/xside.c
```diff
@@ -842,28 +825,151 @@ static void save_clipboard_source_vmname(const char *vmname) {
...
+    if (!file) {
+        perror("Can not open " QUBES_CLIPBOARD_FILENAME ".metadata file");
+        return false;
+    }
+    // Load JSON format
+    while (fgets(line, sizeof(line), file) != NULL) {
+        if (strcmp(line, "") == 0) continue;
+        if (strcmp(line, "{") == 0) continue;
+        if (strcmp(line, "}") == 0) continue;
+        if (! sscanf(line, "\"%[A-Za-z0-9]\":%[\"A-Za-z0-9]s,", key, value)) {
```
marmarek, line 878: Shouldn't this check for `== 2` ?

marmarek: and similarly all below should check sscanf result

alimirjamali:
> The metadata format needs to be documented somewhere. It has many more values than the old one, but also the parser is quite limited and can't understand full JSON - the limitations need to be described somewhere (expected line breaks, one value per line etc)

Exactly. I thought about it as well. Most use cases would be reading the metadata rather than writing to it. I wanted to avoid dependency on any external JSON parsing library for C language to support full JSON. In the end, all of xside.c should be replaced by Demi's rewrite of GUI daemons in Rust?

There is a mention of clipboard files in [qubes-doc](https://github.com/QubesOS/qubes-doc/blob/main/user/how-to-guides/how-to-copy-from-dom0.md?plain=1#L46-L48). But they are for end-user. I still believe that it might be a good place to document the metadata format.

alimirjamali:
> Ok, the message on copy is not really a failure - the first one is expected to not have this file. Can you suppress the error message in this case?

Yes. I guess instead of an error message, `metadata.successful = false` is better. `load_clipboard_metadata` is used only for `xevent_timestamp`. And it would be 0 which is already working.

gui-daemon/xside.c
alimirjamali: > Shouldn't this check for `== 2` ?

Yes

> and similarly all below should check sscanf result

Most of them could be done in one sscanf. Most of them are `%d`. I will do it.

marmarek:
> Yes. I guess instead of an error message, `metadata.successful = false` is better.

TBH, I'm not a fan of mixing content of the loaded file with info about loading process itself. And also, the missing file needs special treatment anyway (compared to other errors on opening the file like permission denied) - missing file is okay on copy, and also kinda expected on some paste (if user didn't copied anything before). But for example permission error or io error are not okay (they should not be treated same as missing file). So, maybe in case of missing file report it via a return value from the function (making it more than a bool) and log message about it only in the paste path? But for other errors log errors directly from the loading function. Or, add a parameter whether missing file message should be logged?

marmarek:
> In the end, all of xside.c should be replaced by Demi's rewrite of GUI daemons in Rust?

One day maybe, but not anytime soon.

Anyway, it's okay to support just subset of JSON (as you said - it's mostly xside who write the file, but there are few where others write it too - for example copying dom0 clipboard via the clipboard widget). But it needs to be documented what that subset is.
You are correct that the current "format" is not documented, I think this would fit into https://www.qubes-os.org/doc/gui/ as other internals of qubes-guid are documented there.

alimirjamali:
> Pasting doesn't work:
> 
> ```
> Failed to parse metadata line: {
> ```

I just realized that sscanf includes `\n` at the end of string.

alimirjamali:
> > Yes. I guess instead of an error message, `metadata.successful = false` is better.
> 
> TBH, I'm not a fan of mixing content of the loaded file with info about loading process itself. And also, the missing file needs special treatment anyway (compared to other errors on opening the file like permission denied) - missing file is okay on copy, and also kinda expected on some paste (if user didn't copied anything before). But for example permission error or io error are not okay (they should not be treated same as missing file). So, maybe in case of missing file report it via a return value from the function (making it more than a bool) and log message about it only in the paste path? But for other errors log errors directly from the loading function. Or, add a parameter whether missing file message should be logged?

The [previous code](https://github.com/QubesOS/qubes-gui-daemon/blob/main/gui-daemon/xside.c#L794-L810) reported non-existent `.xevent` file. The new `get_clipboard_xevent_timestamp` function replaces the old `get_clipboard_file_xevent_timestamp` function. I tried to mimic the old behaviour on errors as much as possible.

I am doing some improvements and will send push them soon.

alimirjamali:
Found the error. In key, value parsing formatting string, I did not include underline as a valid character for key.

alimirjamali:
I am adding a special case if user tries to press ctrl+shift+c in a qube where the clipboard is empty. This can be handled with the new metadata (successful=true, cleared=true, sent_size=0)

alimirjamali:
I made another small adjustment. If `--max-clipboard-size=` command line value is not between 256 & 256000, the qubes-guid should revert to the default. Instead of exiting. Making this more forgiving to users

gui-daemon/xside.c
```diff
@@ -842,28 +825,193 @@ static void save_clipboard_source_vmname(const char *vmname) {
...
+        return false;
+    }
+    // Load JSON format
+    while (fgets(line, sizeof(line), file) != NULL) {
+        if (strlen(line) == 0) continue;
+        if (line[strlen(line) - 1] == '\n') line[strlen(line) - 1] = 0;
+        if (strcmp(line, "") == 0) continue;
+        if (strcmp(line, "{") == 0) continue;
+        if (strcmp(line, "}") == 0) continue;
+        if (sscanf(line, "\"%[A-Za-z0-9_-]\":%[\"A-Za-z0-9_-]s", key, value) != 2) {
```
marmarek, line 881:
```suggestion
        if (sscanf(line, "\"%[A-Za-z0-9_-]\":%[\"A-Za-z0-9_-]", key, value) != 2) {
```
?

gui-daemon/xside.c
```diff
@@ -842,28 +825,193 @@ static void save_clipboard_source_vmname(const char *vmname) {
...
+
+static bool load_clipboard_metadata(struct clipboard_metadata *metadata, bool logging) {
+    FILE *file;
+    char line[256];
+    char key[256] = {0};
+    char value[256] = {0};
+
+    file = fopen(QUBES_CLIPBOARD_FILENAME ".metadata", "r");
+    if (!file) {
+        if (logging)
```
marmarek, line 870: Errors other than ENOENT should result in a logged message even on copy operation.

gui-daemon/xside.h
```diff
@@ -304,6 +305,23 @@ static inline void put_shm_image(
...
+    bool paste_action;          /* if it was a paste action */
+    bool malformed_request;     /* if VM tried to violate its protocol specs */
+    bool oversized_request;     /* if VM tried to send over its limit */
+    bool cleared;               /* if clipboard data is cleared */
+    bool qrexec_clipboard;      /* if clipboard data is from MS-Windows or similar */
+    unsigned int sent_size;     /* number of bytes sent by vmside */
+    unsigned int buffer_size;   /* maximum allowed clipboard size for this vm */
+    uint32_t protocol_version_vmside;    /* inter-vm clipboard GUI protocol version - vmside */
+    uint32_t protocol_version_xside;     /* inter-vm clipboard GUI protocol version - xside */
+};
```
marmarek, line 323: As for documenting the metadata file, some extra explanation is needed for meaning of the fields at different stages (clipboard empty, something copied, something pasted) - especially it's not really obvious that after paste the "vmname" field changes its value.

marmarek (changes requested):

gui-daemon/xside.h
alimirjamali: I updated the explanation and added a link to `https://www.qubes-os.org/doc/gui/` for further information.

gui-daemon/xside.c
```diff
@@ -3807,6 +4014,9 @@ static void usage(FILE *stream)
     fprintf(stream, " --title-name, -T\tprefix window titles with VM name\n");
     fprintf(stream, " --trayicon-mode\ttrayicon coloring mode (see below); default: tint\n");
     fprintf(stream, " --screensaver-name\tscreensaver window name, can be repeated, default: xscreensaver\n");
+    fprintf(stream, " --max-clipboard-size=SIZE\tmaximum clipboard size VM is allowed to send to global clipboard\n");
```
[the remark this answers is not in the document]
alimirjamali: Since it was working with `qvm-start-daemon` synthesized config file, I forgot about the command line. I will add it.

gui-daemon/xside.c
alimirjamali: > Errors other than ENOENT should result in a logged message even on copy operation. Similar comment applies to other messages below messages below. Generally, expected situation should not be reported as an error but others should get logged, and expected case is either fie exists and is correctly parsed, or file doesn't exist yet on the first copy.

I am having problem to properly understand this.
When you advise: `Errors other than ENOENT should result in a logged message even on copy operation`, 
does it mean always log or only if debug mode or verbose mode is enabled?

alimirjamali: @marmarek I applied all the other requested changes. For the error handling & messaging, there are 5 cases within `load_clipboard_metadata` and two cases within `get_clipboard_xevent_timestamp`. I have passed a boolean variable to those function (`g->log_level > 1`).  Maybe it is better to pass it as an integer and log based on different logging levels.

marmarek:
According to the tests the copy operation for 64000 and more fails (with default config and increased limit too). I haven't found any error messages in logs (the metadata has copy_action=1, successful=1, cleared=1).

marmarek:
Hmm, in some cases the 64k case works. But in none of the 4 iterations 200k worked.

alimirjamali:
> Hmm, in some cases the 64k case works. But in none of the 4 iterations 200k worked.

Was the fix I pushed to `gui-agent-linux` 13 hours ago applied to the tests? [This one](https://github.com/QubesOS/qubes-gui-agent-linux/pull/217/commits/79d44348ffd2c6fc5d7445c70509daf2468e75de).

alimirjamali:
Ok. In the updated unittest in [this line](https://github.com/QubesOS/qubes-core-admin/pull/623/files#diff-1c394520b6a434922c7a3e61e6d364ec44edd5f9a47372d2d60ad0583384ebabR585), it sets `gui-max-clipboard-size` feature per vm or per guivm? If it is guivm, it should be `gui-default-max-clipboard-size`

alimirjamali:
In [test_002_clipboard_300k](https://github.com/QubesOS/qubes-core-admin/pull/623/files#diff-1c394520b6a434922c7a3e61e6d364ec44edd5f9a47372d2d60ad0583384ebabR590-R593), does it expect a truncated zero byte `qubes-clipboard.bin`? It should not. The `gui-agent` sends specific string in case of INCR. [This one](https://github.com/QubesOS/qubes-gui-agent-linux/pull/217/files#diff-d2f30726ab5b690e7ac5631b2a30278386f40ada334a45b41787c59535b59333R1063).

I did the above as a temporary solution to INCR. Designing a signal from vmside to xside for the INCR is possible.

alimirjamali:
For `test_002_clipboard_200k_truncated`, it is failing on `AssertionError: '' != 'test-inst-vm1'`. It should not do this. In this case, the new code truncates the `/run/qubes/qubes-clipboard.bin` and sets `/run/qubes/qubes-clipboard.bin.source` to ""

marmarek:
> Ok. In the updated unittest in [this line](https://github.com/QubesOS/qubes-core-admin/pull/623/files#diff-1c394520b6a434922c7a3e61e6d364ec44edd5f9a47372d2d60ad0583384ebabR585), it sets `gui-max-clipboard-size` feature per vm or per guivm? If it is guivm, it should be `gui-default-max-clipboard-size`

It intends to set the per-vm one (for the source VM specifically).



> In [test_002_clipboard_300k](https://github.com/QubesOS/qubes-core-admin/pull/623/files#diff-1c394520b6a434922c7a3e61e6d364ec44edd5f9a47372d2d60ad0583384ebabR590-R593), does it expect a truncated zero byte `qubes-clipboard.bin`? It should not. The `gui-agent` sends specific string in case of INCR. [This one](https://github.com/QubesOS/qubes-gui-agent-linux/pull/217/files#diff-d2f30726ab5b690e7ac5631b2a30278386f40ada334a45b41787c59535b59333R1063).
> 
> I did the above as a temporary solution to INCR. Designing a signal from vmside to xside for the INCR is possible.

Ok, so the 300k test needs an update.

> For `test_002_clipboard_200k_truncated`, it is failing on `AssertionError: '' != 'test-inst-vm1'`. It should not do this. In this case, the new code truncates the `/run/qubes/qubes-clipboard.bin` and sets `/run/qubes/qubes-clipboard.bin.source` to ""

Right, the truncated one indeed should expect empty source. But I'm more worried about non-truncated one failing.

alimirjamali:
> But I'm more worried about non-truncated one failing.

I will look at test result again and see why it is failing. In the meantime, I am working with Marmarta on issue 9029 and will come back to this one very soon. Just for a change (classic c programming could grind mind)

marmarek:
It could be also a race condition in the test. I've added sleep(1) after sending ctrl+shift+c for now, lets see if that changes anything.

alimirjamali:
> According to the tests the copy operation for 64000 and more fails (with default config and increased limit too). I haven't found any error messages in logs (the metadata has copy_action=1, successful=1, cleared=1).

If this fails again, I appreciate if I could have the entire metadata. Is it possible? I am specifically interested on `protocol_version_...`, `buffer_size`. But others will be useful too.

alimirjamali:
> It could be also a race condition in the test. I've added sleep(1) after sending ctrl+shift+c for now, lets see if that changes anything.

I see that `test_002_clipboard_200k` is green. So it was indeed a race condition?

marmarek:
> I see that `test_002_clipboard_200k` is green. So it was indeed a race condition?

Yes, but a different one than I thought initially. Apparently zenity needs a bit of time to process "ctrl+c" when opened with a larger text.

gui-daemon/xside.c
marmarek: Lets leave like this now. Different ENOENT handling technically is a separate improvements not strictly related to this PR.

marmarek (approved):

marmarek:
On Tue, Jun 17, 2025 at 02:19:32AM -0700, Atrate wrote:
> Atrate left a comment (QubesOS/qubes-gui-daemon#150)
> 
> Is this configurable in r4.2 or only in r4.3?

Only R4.3.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

PR #152, opened 2024-10-15
Extend shmoverride init config

marmarek:
Support adding LD_PRELOAD to a parent process too (for example wayland compositor that only later will start Xwayland). See commit messages for details.

Note, this approach does _not_ work together with linking with -z initfirst.

Built on top of https://github.com/QubesOS/qubes-gui-daemon/pull/151


marmarek:
This does _not_ work with kwin. `LD_PRELOAD` is not passed down to children.

PR #153, opened 2024-10-17
Fix bugs in the icon-receiver

qubest:
- Update `guivm-gui-agent` to `guivm`
- When the `icon-reciever` running on guivm other than dom0 the header received is as follow:
  Request coming from appvm to dom0
  ```
  service vm-name keyword adminvm\x00
  ```
  Request coming from appvm to guivm
  ```
  service vm-name\x00
  ```
  The trailing `\x00` cause a bug when trying to get the color of the vm using `qubesadminapi`.

### DIFF
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -342,7 +342,7 @@ class IconReceiver(object):
         try:
             # Parse header from qrexec
             header = await reader.readuntil(b'\0')
-            header_parts = header.decode('ascii').split(' ').restrip('\x00')
+            header_parts = header.decode('ascii').restrip('\x00').split(' ')
             assert len(header_parts) >= 2, header_parts
 
             service_name = header_parts[0]
--- a/window-icon-updater/icon-receiver
+++ b/window-icon-updater/icon-receiver
@@ -388,7 +388,7 @@ def main():
     args = parser.parse_args()
 
     if not args.force:
-        if (not os.path.exists('/var/run/qubes-service/guivm-gui-agent') and
+        if (not os.path.exists('/var/run/qubes-service/guivm') and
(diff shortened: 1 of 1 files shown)


marmarek (approved):

window-icon-updater/icon-receiver
```diff
@@ -342,7 +342,7 @@ class IconReceiver(object):
         try:
             # Parse header from qrexec
             header = await reader.readuntil(b'\0')
-            header_parts = header.decode('ascii').split(' ')
+            header_parts = header.decode('ascii').split(' ').restrip('\x00')
```
marmarek, line 345: List doesn't have `restrip()` method. Did you mean `.rstrip()` just before `.split()` ?

marmarek (changes requested):

window-icon-updater/icon-receiver
qubest: It is a typo sorry.

Yep, the rstrip should be added before split.
Or at header_parts[1].rstrip('\x00')

marmarek: Better before split, as this is terminator of this whole header, not a specific part

window-icon-updater/icon-receiver
```diff
@@ -342,7 +342,7 @@ class IconReceiver(object):
         try:
             # Parse header from qrexec
             header = await reader.readuntil(b'\0')
-            header_parts = header.decode('ascii').split(' ')
+            header_parts = header.decode('ascii').restrip('\x00').split(' ')
```
marmarek, line 345:
```suggestion
            header_parts = header.decode('ascii').rstrip('\x00').split(' ')
```

marmarek (changes requested):

marmarek (approved):

PR #155, opened 2024-11-08
Replace U+2026 with ... if utf8 is disabled

alimirjamali:
Agent side should have replaced the end of very long window titles to
U+2026 (Horizontal Ellipsis). But the Daemon side might have been
configured to disallow UTF8 characters. Replace that character with ...

Supplements: https://github.com/QubesOS/qubes-gui-agent-linux/pull/227
Main issue: https://github.com/QubesOS/qubes-issues/issues/5986

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -3068,8 +3068,15 @@ static void handle_wmname(Ghandles * g, struct windowdata *vm_window)
     read_struct(g->vchan, untrusted_msg);
     /* sanitize start */
     untrusted_msg.data[sizeof(untrusted_msg.data) - 1] = 0;
+    // If the agent has changed the end of very long title to U+2026 but utf8 is disabled
+    if (!g->allow_utf8_titles
+        && strlen(untrusted_msg.data) > 3
+        && strcmp(untrusted_msg.data + strlen(untrusted_msg.data) - 3, "\xE2\x80\xA6") == 0) {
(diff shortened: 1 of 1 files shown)


gui-daemon/xside.c
```diff
@@ -2912,6 +2912,12 @@ static void handle_wmname(Ghandles * g, struct windowdata *vm_window)
     read_struct(g->vchan, untrusted_msg);
     /* sanitize start */
     untrusted_msg.data[sizeof(untrusted_msg.data) - 1] = 0;
+    /* change the end of very long window titles to ... */
+    if (strlen(untrusted_msg.data) == (sizeof(untrusted_msg.data) - 1)) {
```
marmarek, line 2916: Shouldn't this be done on the agent side instead? It knows where the limit is too, and may have more info (especially, if it really was truncated or maybe just at the limit).

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: > Shouldn't this be done on the agent side instead? It knows where the limit is too, and may have more info (especially, if it really was truncated or maybe just at the limit).

I did it on the agent side at the beginning. I am going to change this PR to draft and do it properly, writing the main patch for both agent and xside to have the length increased (from 128 to 4096) as well as appending `...` if necessary.

PR #156, opened 2024-11-08
Add `_NET_WM_PING` support

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/6950

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -64,6 +64,7 @@
 #include "trayicon.h"
 #include "shm-args.h"
 #include "util.h"
+#include "unistd.h"
 #include <qubes/pure.h>
 
 /* Supported protocol version */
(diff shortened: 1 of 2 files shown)


marmarek:
For this to be more useful it would make sense to set also `_NET_WM_PID`.

alimirjamali:
> For this to be more useful it would make sense to set also `_NET_WM_PID`.

Sure. But there are issues. The [specification](https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html#id-1.6.14) says:
> If _NET_WM_PID is set, the ICCCM-specified property WM_CLIENT_MACHINE MUST also be set.

And if `WM_CLIENT_MACHINE` is set, the existing Qubes OS [patch](https://github.com/QubesOS/qubes-desktop-linux-xfce4-xfwm4/blob/main/0001-Qubes-decoration.patch#L134) to xfce-xfwm4 will use it and the window title would be:
`[_QUBES_VMNAME] WM_NAME (WM_CLIENT_MACHINE)`
Result would be something like `[personal] user - Thunar (personal)` which is ugly. The 2nd Qube name in parentheses is redundant. This happens at least on XFCE. I faced this issue while working on [#9554](https://github.com/QubesOS/qubes-issues/issues/9554).

Even without the `_NET_WM_PID`, xfce is able to detect windows not responding to ping and successfully kill the related GUID process (I tested it and confirmed that). I am not sure how it does it without `_NET_WM_PID`. Maybe it has its own hidden property.

I am in the process of study and experimenting with implementation of the missing EWMH and Xorg specs.

marmarek:
> And if `WM_CLIENT_MACHINE` is set, the existing Qubes OS [patch](https://github.com/QubesOS/qubes-desktop-linux-xfce4-xfwm4/blob/main/0001-Qubes-decoration.patch#L134) to xfce-xfwm4 will use it and the window title would be:
> `[_QUBES_VMNAME] WM_NAME (WM_CLIENT_MACHINE)`

Only if WM_CLIENT_MACHINE is set to something else than local host (dom0 or guivm name). In fact, setting to WM_CLIENT_MACHINE is set to the actual VM name, then killing hung process won't work.

alimirjamali:
> Only if WM_CLIENT_MACHINE is set to something else than local host (dom0 or guivm name). In fact, setting to WM_CLIENT_MACHINE is set to the actual VM name, then killing hung process won't work.

Very well. I will write it.

PR #157, opened 2024-11-11
Add `_NET_WM_NAME` & `_NET_WM_ICON_NAME` support

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/9571

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -600,6 +600,9 @@ static void intern_global_atoms(Ghandles *const g) {
         { &g->wm_user_time, "_NET_WM_USER_TIME" },
         { &g->wmDeleteMessage, "WM_DELETE_WINDOW" },
         { &g->net_supported, "_NET_SUPPORTED" },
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -2925,7 +2925,17 @@ static void handle_wmname(Ghandles * g, struct windowdata *vm_window)
     Xutf8TextListToTextProperty(g->display, list, 1, XUTF8StringStyle,
                     &text_prop);
     XSetWMName(g->display, vm_window->local_winid, &text_prop);
+    XChangeProperty(g->display, vm_window->local_winid,
+        XInternAtom(g->display, "_NET_WM_NAME", False),
```
marmarek, line 2929: Don't call it over and over, cache it in ghandles. See intern_global_atoms(), and same for the other calls below.

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: Done

PR #158, opened 2024-11-12
Clear global clipboard after specified seconds

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/6641

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -880,6 +880,7 @@ static void save_clipboard_metadata(struct clipboard_metadata *metadata) {
     fprintf(file, "\"buffer_size\":%d,\n", metadata->buffer_size);
     fprintf(file, "\"protocol_version_xside\":%d,\n", metadata->protocol_version_xside);
     fprintf(file, "\"protocol_version_vmside\":%d\n", metadata->protocol_version_vmside);
+    fprintf(file, "\"timeout\":%d\n", metadata->timeout);
     // other key,value pairs could be added if needed in future
     fprintf(file, "}\n");
     fclose(file);
@@ -959,6 +960,8 @@ static bool load_clipboard_metadata(struct clipboard_metadata *metadata, bool lo
             metadata->protocol_version_vmside = dummy;
         } else if (strcmp(key, "protocol_version_xside") == 0) {
             metadata->protocol_version_xside = dummy;
+        } else if (strcmp(key, "timeout") == 0) {
+            metadata->timeout = dummy;
         }
     }
     fclose(file);
@@ -1219,6 +1222,13 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     metadata.protocol_version_vmside = g->protocol_version;
     metadata.protocol_version_xside = PROTOCOL_VERSION(
                     PROTOCOL_VERSION_MAJOR, PROTOCOL_VERSION_MINOR);
+    metadata.timeout = g->clipboard_timeout;
+
+    // if a clipboard timeout child is pending, terminate it and clear its PID.
+    if (g->clipboard_timeout_pid) {
+        kill(g->clipboard_timeout_pid, SIGTERM);
+        g->clipboard_timeout_pid = 0;
+    }
 
     if (g->log_level > 0)
         fprintf(stderr, "handle_clipboard_data, len=0x%x\n",
@@ -1301,6 +1311,42 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     save_clipboard_file_xevent_timestamp(g->clipboard_xevent_time);
     metadata.successful = true;
     save_clipboard_metadata(&metadata);
+
+    if (g->clipboard_timeout > 0) {
+        struct stat st0, st1;
+        if (lstat(QUBES_CLIPBOARD_FILENAME, &st0) == -1) {
+            perror("Can not get clipboard data file status");
+            exit(1);
+        }
+        switch (g->clipboard_timeout_pid = fork()) {
+            case -1:
+                perror("fork");
+                exit(1);
+            case 0:
+                if (sleep(g->clipboard_timeout) != 0) {
+                    perror("Clipboard clearing timeout was interrupted");
+                    _exit(1);
+                }
+                inter_appviewer_lock(g, 1);
+                if (lstat(QUBES_CLIPBOARD_FILENAME, &st1) == -1) {
+                    perror("Can not get clipboard data file status");
+                    _exit(1);
+                }
+                if (st0.st_mtime == st1.st_mtime) {
+                    old_umask = umask(0007);
+                    metadata.copy_action = false;
+                    metadata.timeout = g->clipboard_timeout;
+                    clear_clipboard(&metadata);
+                    umask(old_umask);
+                    if (g->log_level > 1)
+                        fprintf(stderr, "Clipboard cleared after %d seconds\n",
+                            g->clipboard_timeout);
+                }
+                inter_appviewer_lock(g, 0);
+                _exit(1);
+        }
+    }
+
 error:
     umask(old_umask);
     inter_appviewer_lock(g, 0);
@@ -4494,6 +4540,8 @@ static void load_default_config_values(Ghandles * g)
     g->paste_seq_mask = ControlMask | ShiftMask;
     g->paste_seq_key = XK_v;
     g->clipboard_buffer_size = DEFAULT_CLIPBOARD_BUFFER_SIZE;
+    g->clipboard_timeout = 0;
+    g->clipboard_timeout_pid = 0;
     g->allow_fullscreen = 0;
     g->override_redirect_protection = 1;
     g->startup_timeout = 45;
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -1274,6 +1274,38 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
     save_clipboard_metadata(&metadata);
+
+    if (g->clipboard_timeout > 0) {
+        switch (fork()) {
+            case -1:
+                perror("fork");
+                exit(1);
+            case 0:
+                struct stat st0, st1;
+                if (lstat(QUBES_CLIPBOARD_FILENAME, &st0) == -1) {
```
marmarek, line 1285: The original stat should be done still under the lock (which means moving it before fork()), otherwise you have a race condition here.

marmarek (changes requested):

marmarek:
> ast second thoughts: If we store `g->clipboard_timeout` in `metadata.sent_size`, the clipboard systray widget could show a better pop-up message. Something like:

Interesting idea. The copy notification could also say something like "... and will be cleared in XX seconds".

alimirjamali:
> Interesting idea. The copy notification could also say something like "... and will be cleared in XX seconds".

In this case, I think it is better to have dedicated key for it in the JSON (like) metadata. Easy to implement. I will do it.

gui-daemon/xside.c
```diff
@@ -1274,6 +1274,38 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     save_clipboard_file_xevent_timestamp(g->clipboard_xevent_time);
     metadata.successful = true;
     save_clipboard_metadata(&metadata);
+
+    if (g->clipboard_timeout > 0) {
+        switch (fork()) {
```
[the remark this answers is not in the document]
alimirjamali: > Then, the parent need to handle SIGCHLD and call wait() on child exit, otherwise you'll leave zombie processes (alternative could be setting SIGCHLD handler to ignore).

I am not very familiar with this and I should study `wait_for_vchan_or_argfd` code. What happens if the parent wants to wait on system shutdown? For example if user set `cllipboard_timeout` to 30 minutes and did a copy operation 1 minute ago?

gui-daemon/xside.c
```diff
@@ -1274,6 +1274,38 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+            case -1:
+                perror("fork");
+                exit(1);
+            case 0:
+                struct stat st0, st1;
+                if (lstat(QUBES_CLIPBOARD_FILENAME, &st0) == -1) {
+                    perror("Can not get clipboard data file status");
+                    exit(1);
+                }
+                sleep(g->clipboard_timeout);
```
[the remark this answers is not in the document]
alimirjamali: done

gui-daemon/xside.c
```diff
@@ -1274,6 +1274,38 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+                if (lstat(QUBES_CLIPBOARD_FILENAME, &st1) == -1) {
+                    perror("Can not get clipboard data file status");
+                    exit(1);
+                }
+                if (st0.st_mtime == st1.st_mtime) {
+                    inter_appviewer_lock(g, 1);
```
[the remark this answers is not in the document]
alimirjamali: done

gui-daemon/xside.c
alimirjamali: done

alimirjamali:
Further notes:
We have to consider the qrexec clipboard case. And what to do for Microsoft Windows. I did not bother for the time while the QWT support is not officially restored. I will work on it afterwards. As there are many corner cases to cover.

gui-daemon/xside.c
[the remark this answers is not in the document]
alimirjamali: I think I got it how to handle this properly. Each main GUID process should not have more than one active clipboard cleanup child (more is meaningless). And its PID could be stored in ghandles. If a new cleanup child is to be created, the old one has to be killed (via SIGTERM). And the main loop could also kill any existing one via SIGTERM.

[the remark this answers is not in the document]
marmarek: > I am not very familiar with this and I should study `wait_for_vchan_or_argfd` code

Actually, wait a moment, https://github.com/QubesOS/qubes-gui-daemon/pull/149 already introduces the timeout parameter.

gui-daemon/xside.c
```diff
@@ -4800,6 +4850,8 @@ int main(int argc, char **argv)
                 busy = 1;
             }
         } while (busy);
+        if (ghandles.clipboard_timeout_pid)
```
marmarek, line 4853: If `fork()` failed, this might be -1. Sending signal to PID -1 is a _very_ bad idea (I'll let you check in docs what it does...). I know the current code exits when that fork fails, but IMO better to check for `pid > 0` just in case it changes in the future.

gui-daemon/xside.c
```diff
@@ -1274,6 +1278,45 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
     save_clipboard_file_xevent_timestamp(g->clipboard_xevent_time);
     metadata.successful = true;
     save_clipboard_metadata(&metadata);
+
+    if (g->clipboard_timeout > 0) {
+        struct stat st0, st1;
+        // if a clipboard timeout child is pending, terminate it.
+        if (g->clipboard_timeout_pid)
+            kill(g->clipboard_timeout_pid, SIGTERM);
```
marmarek, line 1286: When you'll be fixing zombie issue, clear this variable too, to avoid killing unrelated processes (that happen to reuse this PID later).

gui-daemon/xside.c
```diff
@@ -1274,6 +1278,45 @@ static void handle_clipboard_data(Ghandles * g, unsigned int untrusted_len)
...
+        if (lstat(QUBES_CLIPBOARD_FILENAME, &st0) == -1) {
+            perror("Can not get clipboard data file status");
+            exit(1);
+        }
+        switch (g->clipboard_timeout_pid = fork()) {
```
marmarek, line 1291: but if you clear the pid variable on SIGCHLD, depending on the execution order, it may override the new PID instead of the old one, so be careful.

marmarek (changes requested):

gui-daemon/xside.c
marmarek: (those comments apply only if you stick with a child process for this; moving cleanup to the main loop avoids those issues completely)

alimirjamali: > (those comments apply only if you stick with a child process for this; moving cleanup to the main loop avoids those issues completely)

I moved it to top of the `handle_clipboard_data` function, before even checking for `g->clipboard_tiemout > 0` check. Hope this is good.

gui-daemon/xside.c
alimirjamali: Oh. Now I see it. In this case, the child process might have completed its job and ended successfully, clearing the clipboard. And a totally unrelated process might have the same PID.

gui-daemon/xside.c
DemiMarie: Instead of using a subprocess, I recommend using a timer integrated with the event loop.  This is much cleaner and avoids all of the problems you are having with subprocesses.  Using `fork()` for concurrency is something Qubes OS does a lot, but it’s generally considered bad practice nowadays.  Newer programs usually an event loop.

The two ways to deal with timeouts are to either use a timeout for `poll()` (updated each call) or a Linux timerfd.

DemiMarie (changes requested):
Please avoid using `fork()`.  Managing child processes creates a lot of complexity and isn’t needed here.

marmarek:
Actually, there is one benefit of using `fork()` here. It allows to clear the clipboard even if qube is shutdown in the meantime. @marmarta said it would be confusing if clearing the clipboard would depend on the source qube remain running long enough.

marmarek:
But also, the child process could be used just for the case of early qube shutdown, and otherwise use the event loop.

DemiMarie:
> But also, the child process could be used just for the case of early qube shutdown, and otherwise use the event loop.

Alternatively the process could keep running even if the qube shut down, or just wipe the clipboard on qube shutdown.

@marmarta: what do you think of wiping the clipboard when the qube that pasted into it is shut down?

marmarek:
> or just wipe the clipboard on qube shutdown.

I already asked, see my comment above.



> Alternatively the process could keep running even if the qube shut down

That may be messy to do. You'll need to do most cleanup (destroy windows, unlink pidfile etc) but not the actual exit. But I don't have much better idea (keeping a child running is quite similar).

alimirjamali:
I have been thinking about this. The way we are trying to do it is wrong. We do not need multiple clipboard clearing processes or even one per GUI  daemon instance. We only need one per entire GUIVM since there is only one instance of global clipboard.

On a new copy action, the GUI daemon instance handling the copy action should check if the mentioned clearing process exists and terminate it gracefully before overwriting the global clipboard. Then launch a new one if needed. This program is not necessarily the GUI daemon itself. It does not even require connection to X Window.

gui-daemon/xside.c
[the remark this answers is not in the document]
marmarek: Ok, now that #149 is merged, wait_for_vchan_or_argfd has a timeout argument.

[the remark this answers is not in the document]
alimirjamali: > Ok, now that #149 is merged, wait_for_vchan_or_argfd has a timeout argument.

Good. I will resume working on it. Writing the weekly newsletter already exhausted my energy for tonight. I will resume working tomorrow.

PR #163, opened 2025-05-20
Assure Popup Windows (override-redirect) open over parent

alimirjamali:
The problem and solution is reported by `vx-sec` on  Github. It was found on Xmonad WM. The fix simply raises the override-redirect window to top, flushes display events and the restack windows for "Keep-on-top" windows to work properly

Also tested on the default XFCE on r4.3

resolves: https://github.com/QubesOS/qubes-issues/issues/9935

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -3357,8 +3357,8 @@ static bool should_keep_on_top(Ghandles *g, Window window) {
 
 
 /* Move a newly mapped override_redirect window below windows that need to be
- * kept on top, i.e. screen lockers. */
-static void restack_windows(Ghandles *g, struct windowdata *vm_window)
+ * kept on top, i.e. screen lockers. Returns new index (-1 == top of all) */
+static int restack_windows(Ghandles *g, struct windowdata *vm_window)
 {
     Window root;
     Window parent;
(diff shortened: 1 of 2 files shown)


gui-daemon/xside.c
```diff
@@ -3466,6 +3466,10 @@ static void handle_map(Ghandles * g, struct windowdata *vm_window)
     (void) XMapWindow(g->display, vm_window->local_winid);
 
     if (vm_window->override_redirect) {
+        // Ensure override redirect window is raised after mapping
+        XRaiseWindow(g->display, vm_window->local_winid);
```
marmarek, line 3470: This solves the original problem, but introduces another: that popup may flash on top of a screenlocker. Not desirable, especially for notifications that may contain sensitive info.
Better to modify `restack_windows` to move it above other windows, but below the screenlocker. Or at the very least, make `restack_windows` return info whether the are any windows that should be on top, and only do `XRaiseWindow` if there are none.

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: This is done. I implemented the 2nd suggested approach.

marmarek (approved):

PR #165, opened 2025-06-25
Notify users on clipboard paste denied by policies

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/9978

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -1377,10 +1377,10 @@ static bool write_all(int socket, void *buf, size_t len) {
     return true;
 }
 
-static int evaluate_clipboard_policy_dom0(Ghandles *g,
+static bool evaluate_clipboard_policy_dom0(Ghandles *g,
         const char *const source_vm) {
     const int sockfd = socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0);
-    int status = 0;
+    bool status = false;
     if (sockfd < 0) {
         perror("socket");
         return false;
@@ -1458,6 +1458,7 @@ static int evaluate_clipboard_policy(Ghandles * g) {
     int fd;
     ssize_t len;
     char source_vm[255];
+    bool result;
 
     fd = open(QUBES_CLIPBOARD_FILENAME ".source", O_RDONLY);
     if (fd < 0)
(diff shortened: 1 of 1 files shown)


gui-daemon/xside.c
```diff
@@ -1569,6 +1569,7 @@ static int is_special_keypress(Ghandles * g, const XKeyEvent * ev, XID remote_wi
             return 1;
         }
         if (!evaluate_clipboard_policy(g)) {
+            show_message(g, "ERROR", "Pasting to this qube is denied by policies", NOTIFY_EXPIRES_DEFAULT);
```
marmarek, line 1572: It isn't necessarily about "pasting to this qube", it's also about from where the clipboard is. So, maybe include the source name in the message too?
Alternatively, make qrexec notify the user, as it's done for normal services. It's mostly about this line: https://github.com/QubesOS/qubes-core-qrexec/blob/main/qrexec/tools/qrexec_policy_exec.py#L405. 
While theoretically enabling notifications for all "just_evaluate" (denied) calls is not a good idea, in practice that "just_evaluate" mode is currently used only for clipboard operations, so maybe it's still one-line change?

marmarek (changes requested):

gui-daemon/xside.c
alimirjamali: > It isn't necessarily about "pasting to this qube", it's also about from where the clipboard is. So, maybe include the source name in the message too?

This is done

> While theoretically enabling notifications for all "just_evaluate" (denied) calls is not a good idea, in practice that "just_evaluate" mode is currently used only for clipboard operations, so maybe it's still one-line change?

Maybe in future, it might be necessary to `just_evaluate` for other operations? So lets do not burn options?

alimirjamali:
@marmarek Something is confusing. I have:
```
qubes.ClipboardPaste * @anyvm untrusted deny
```
and it is still possible clipboard from `dom0` to `untrusted`. And I believe it is not related to this PR.

marmarek:
> `qubes.ClipboardPaste * @anyvm untrusted deny`

`@anyvm` matches everything except dom0.

marmarek (approved):

PR #172, opened 2025-10-10
Don't event-buffer ConfigureNotify, MapNotify, and ReparentNotify events

### LINKED ISSUE #10286: Slight flicker when opening Qt application menus in qubes with gui-events-max-delay feature set
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release

R4.3

### Brief summary

When using Qt applications, qubes which have the `gui-events-max-delay` feature set to a non-zero value may exhibit flickering when opening a menu in an application's menubar. The flicker doesn't occur 100% of the time, and it isn't massive, but it's definitely there. For unknown reasons this flicker does *not* appear to occur in GTK applications like Thunar and xfce4-terminal, but does occur in multiple Qt applications (qterminal, pcmanfm-qt).

### Steps to reproduce

* Launch xfce4-terminal in an `anon-whonix` AppVM (since it will have `gui-events-max-delay` set to `100` by default).
* Install `qterminal`.
* Run `qterminal`.
* Click on one of the items in the application menu bar, such as "File".

### Expected behavior

Menu should immediately appear and not flicker.

### Actual behavior

Menu appears, then seems to very briefly vanish, then re-appear (flicker).

### Additional information

So far I've reproduced this on Whonix-Workstation 18 and Debian 13. Setting `gui-events-max-delay` to 0 causes the flicker to stop.

ArrayBolt3:
When event buffering is enabled for a qube, previously Qt applications would exhibit some strange flickering / element resizing when opening application menus. This appears to be because ConfigureNotify events were being delayed, causing Qt to draw elements of the user interface with incorrect size and position information. It would then likely have to redraw the menu with the correct information once that information was available to it.

ConfigureNotify, MapNotify, and ReparentNotify events all look like they can be passed through without negative effects on anonymity, so make them exempt from event buffering. This appears to resolve the flickering issue.

Fixes: https://github.com/QubesOS/qubes-issues/issues/10286

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -2815,7 +2815,16 @@ static void process_xevent(Ghandles * g)
     XEvent event_buffer;
     XNextEvent(g->display, &event_buffer);
     if (g->ebuf_max_delay > 0) {
-        ebuf_queue_xevent(g, event_buffer);
+        switch (event_buffer.type) {
+        case ConfigureNotify:
+        case ReparentNotify:
+        case MapNotify:
+            process_xevent_core(g, event_buffer);
+            break;
+        default:
+            ebuf_queue_xevent(g, event_buffer);
+            break;
+        }
     } else {
         process_xevent_core(g, event_buffer);
     }


ArrayBolt3:
@marmarek The flickering issue can be reproduced without this PR in both QTerminal and PCManFM-Qt. The reproduction steps are in the linked bug report, it sometimes takes more than one try before the steps succeed in reproducing the issue, but I'm able to reproduce it reliably. I was not able to reproduce it after applying these changes.

marmarek (approved):

PR #173, opened 2025-11-12
Re-dock tray icons when tray manager is restarted

### LINKED ISSUE #2242: System tray icons disappearing
#### Qubes OS version (e.g., `R3.1`):

`R3.2-rc2`
#### Problem description:

Multiple users have reported system tray icons "disappearing" while using the i3 window manager:
- https://groups.google.com/d/topic/qubes-users/pRE0yr_EnIQ/discussion
- https://groups.google.com/d/topic/qubes-users/nN7JWaQggZQ/discussion

And under Xfwm4:
- https://github.com/QubesOS/qubes-issues/issues/2242#issuecomment-242303198

And under KDE:
- #2268

---
#### Possibly related issues:
#2216, #2264, #2268

marmarek:
Listen for MANAGER events about tray selection manager and then dock
icons into the new dock area. New tray manager is supposed to sent such
message after taking over the function of the old one. There is no need
for undocking from the old one, as that's automatic - at the point the
message is sent, old manager must be already destroyed and all its
resources released. See https://tronche.com/gui/x/icccm/sec-2.html#s-2.8
for detailed interaction.

This fixes tray icons in case of dock restart (rare on Xfce, but
frequent on i3 - every config reload).

For this work work reliably on i3, it needs also ignoring failure to
send the dock message - dock may be restarted twice, introducing race
condition (sending message to already destroyed dock). The code in fact
ignores BadWindow error of any XSendEvent call, but in all the cases
it's the same situation.

Fixes QubesOS/qubes-issues#2242

### DIFF
--- a/gui-daemon/xside.c
+++ b/gui-daemon/xside.c
@@ -259,6 +259,10 @@ static int x11_error_handler(Display * dpy, XErrorEvent * ev)
         fprintf(stderr, "  someone else already destroyed this window, ignoring\n");
         return 0;
     }
+    if (ev->request_code == X_SendEvent && ev->error_code == BadWindow) {
+        fprintf(stderr, "  likely window manager restart in progress, ignoring\n");
+        return 0;
+    }
     /* Permit XGetWindowAttributes errors, as long as they're not for root_win */
     if (ev->request_code == X_GetWindowAttributes &&
         ev->error_code == BadWindow &&
@@ -619,6 +623,7 @@ static void intern_global_atoms(Ghandles *const g) {
         { &g->net_wm_name, "_NET_WM_NAME" },
         { &g->net_wm_icon_name, "_NET_WM_ICON_NAME" },
         { &g->utf8_string, "UTF8_STRING" },
+        { &g->manager, "MANAGER" },
     };
     Atom labels[QUBES_ARRAY_SIZE(atoms_to_intern)];
     const char *names[QUBES_ARRAY_SIZE(atoms_to_intern)];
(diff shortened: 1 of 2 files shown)


marmarek:
@ben-grande nm-applet survived: https://openqa.qubes-os.org/tests/159348#step/i3_startup/14
Whonix icon didn't, not sure why, but in local tests it looks like reload gets it back (with this PR).

ben-grande:
> @ben-grande nm-applet survived: https://openqa.qubes-os.org/tests/159348#step/i3_startup/14 Whonix icon didn't, not sure why, but in local tests it looks like reload gets it back (with this PR).

I understand it is a different issues about items being blank, but it may help explain the issue:

- https://forum.qubes-os.org/t/some-kde-system-tray-icons-are-blank/23526/4

> Let’s check the source code for the working applets:
> 
> sdwdate-gui: [sdwdate-gui/usr/lib/python3/dist-packages/sdwdate_gui/sdwdate_gui.py at master · Kicksecure/sdwdate-gui · GitHub](https://github.com/Kicksecure/sdwdate-gui/blob/master/usr/lib/python3/dist-packages/sdwdate_gui/sdwdate_gui.py)
> VLC: [vlc/modules/gui/qt/dialogs/systray/systray.cpp at master · videolan/vlc · GitHub](https://github.com/videolan/vlc/blob/master/modules/gui/qt/dialogs/systray/systray.cpp#L42)
> 
> Both are using Qt’s `QSystemTrayIcon` class. And `nm-applet` used for network manager is GTK3. So we can assume from here that GTK is buggy but not Qt.

Maybe test with other pre-intsalled applications that can appear on the tray such as KeepassXC (Qt, has to enable systray in the app) and volumeicon from sys-audio?

- Qt: SDWDate, KeepassXC
- GTK: nm-applet, volumeicon

Then we would have a larger sample to see if it is related to the framework.

marmarek:
> I understand it is a different issues

It's indeed a different issue, please put it in appropriate place (https://github.com/QubesOS/qubes-issues/issues/9945 ?)

ben-grande:
> > I understand it is a different issues
> 
> It's indeed a different issue

My comment was intended to help test if the tray icons disappear for Qt but not for GTK. Not pushing to fix the blank issue in this PR.

marmarek:
> * Qt: SDWDate, KeepassXC

Neither of those survive.

> * GTK: nm-applet, volumeicon

But those do.

So, indeed looks to be toolkit related

marmarek:
So, this solves only half of the problem, but I'm going to merge it anyway. The other half can be solved in another PR.

PR #174, opened 2025-12-18
Allow user to set qube audio volume

### LINKED ISSUE #2724: Disable speaker output for Whonix WS-based AppVMs by default
To prevent them from communicating to other mic-equipped devices and deanonymizing the user. (Note we do not expose mic to any AppVM by default since forever.)

Seems like we can easily do that during installation via Salt config. But maybe consider also some runtime check to ensure this also (in Qubes 4.x)?

/cc @adrelanos

alimirjamali:
This is the `pacat-simple-vchan` core part of the patch to introduce a `-V volume` command line option. It is used by `qvm-start-daemon` and salt formulas.

resolves: https://github.com/QubesOS/qubes-issues/issues/2724

### DIFF
--- a/pulse/pacat-simple-vchan.c
+++ b/pulse/pacat-simple-vchan.c
@@ -406,7 +406,7 @@ static void send_rec_data(pa_stream *s, struct userdata *u, bool discard_overrun
     pa_stream_drop(s);
 }
 
-/* This is called whenever new data may is available */
+/* This is called whenever new data may be available */
 static void stream_read_callback(pa_stream *s, size_t length, void *userdata) {
     struct userdata *u = userdata;
 
@@ -730,6 +730,8 @@ static bool connect_disconnect_rec_stream(
 static void context_state_callback(pa_context *c, void *userdata) {
     struct userdata *u = userdata;
     pa_stream_flags_t flags = 0;
+    pa_cvolume cvolume = {1, {u->volume}};
+    pa_cvolume *vol = NULL;
 
     assert(c);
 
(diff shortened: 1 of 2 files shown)


marmarek:
(moving discussion from https://github.com/QubesOS/qubes-core-admin-client/pull/405)

> How should it be dealt with DispVMs based on whonix-workstation-xx-dvm? Shouldn't they be muted by default?

If something should not have audio at all, setting audiovm to none is the proper way. Nowadays dynamic change of audiovm should work, so if one wishes, they can enable audiovm later too.

> For those we do not want to set and we want pulse to remember the previous setting, I could easily adjust `pacat-simple-vchan` to pass NULL (like before) to `pa_stream_connect_playback` if `audio-volume` feature is not set. So the previous volume is preserved.

That still won't work as one would expect. `pa_stream_connect_playback`  is also called when pulseaudio/pipewire is restarted (which may also happen on its own in some cases), and your implementation would reset the volume in such case. And finally, "volume" and "mute" are two separate things. While you can emulate mute with setting volume to 0, it's not the proper way - for example it's annoying to un-mute and have it the same volume as other qubes.

alimirjamali:
> > How should it be dealt with DispVMs based on whonix-workstation-xx-dvm? Shouldn't they be muted by default?
> 
> If something should not have audio at all, setting audiovm to none is the proper way. Nowadays dynamic change of audiovm should work, so if one wishes, they can enable audiovm later too.

For the DispVMs based on `whonix-workstation-xx-dvm`, I assumed users may want them to be muted at start rather than having `audiovm` completely removed? While dynamic set of `audiovm` for such DispVMs is possible,  I guess it is not that convenient?

> That still won't work as one would expect. `pa_stream_connect_playback` is also called when pulseaudio/pipewire is restarted (which may also happen on its own in some cases), and your implementation would reset the volume in such case. And finally, "volume" and "mute" are two separate things. While you can emulate mute with setting volume to 0, it's not the proper way - for example it's annoying to un-mute and have it the same volume as other qubes.

For this issue, I believe I can do it before the `g_main_loop_run` to avoid resetting the volume in the mentioned cases. And the `audio-volume` could be dual purpose. An integer value between 0 to 150 to set the volume and a `mute` string to actually mute rather than setting the volume?

marmarek:
> For the DispVMs based on `whonix-workstation-xx-dvm`, I assumed users may want them to be muted at start rather than having `audiovm` completely removed? While dynamic set of `audiovm` for such DispVMs is possible, I guess it is not that convenient?

True, currently requires CLI, while volume/mute can be changed via standard volume control. Kinda related to https://github.com/QubesOS/qubes-issues/issues/10449

> For this issue, I believe I can do it before the `g_main_loop_run` to avoid resetting the volume in the mentioned cases. 

I don't think so, stream objects are not created at this point yet.

> And the `audio-volume` could be dual purpose. An integer value between 0 to 150 to set the volume and a `mute` string to actually mute rather than setting the volume?

This might be okay (but honestly, I still think using anything other than `mute` here would lead to issues).

alimirjamali:
> > And the `audio-volume` could be dual purpose. An integer value between 0 to 150 to set the volume and a `mute` string to actually mute rather than setting the volume?
> 
> This might be okay (but honestly, I still think using anything other than `mute` here would lead to issues).

I amended the patch to implement both `volume` value and `mute` option.

> That still won't work as one would expect. `pa_stream_connect_playback` is also called when pulseaudio/pipewire is restarted (which may also happen on its own in some cases), and your implementation would reset the volume in such case

I implemented some measures to set volume or mute only at the first call to `pa_stream_connect_playback`. I have tested it and it works.

marmarek (approved):

